PR #4, opened 2015-09-27
gradle build system

ibit:
* crashlytics init is commented out
* built for API 22 with the latest Google Play Services


smarx:

@ibit, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

app/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    compile fileTree(dir: 'libs', include: ['*.jar'])
```
RX14: There aren't any libs not on maven so this isn't needed.

RX14:
I think we should merge the best bits of this PR and RX14/zulip-android, because @AndyScherzinger has sent a PR with some fixes for android 5: RX14/zulip-android#1.

On the other hand, the person who made this PR probably has developed on android before :P

app/app.iml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
RX14: Should this not be gitignored and generated by IDEA?

AndyScherzinger: imho yes, since the iml is intellij's project file which gets generated automatically during project import.

ibit:
@AndyScherzinger I added your changes manually to this PR.
@smarx signed

RX14:
If this looks good to everyone else, I guess it just needs to be squashed into one or 2 commits.

AndyScherzinger:
Sounds good to me even though I couldn't test drive the PR yet, but overall it is a decision of the zulip team since we bumped the min version to Android 4.0.3 (which imho is fine).

se-bastiaan:
Who is in charge of this repo anyways? I see merges in the server repo, but no one is responding here.

timabbott:
Sorry about that!  The Zulip team only had one Android developer who's very busy right now with other work and so hasn't been able to be as responsive.  We're working on finding time for someone with time to review and merge the changes and do a production build; definitely keep the patches coming!

app/build.gradle
```diff
@@ -0,0 +1,43 @@
+apply plugin: 'com.android.application'
+
+repositories {
+    mavenCentral()
+    maven { url 'http://download.crashlytics.com/maven' }
```
lfaraone: I'd prefer we not use Maven repositories that aren't served over HTTPS.

It looks like Crashlytics doesn't provide a Maven repository over HTTPS, but [Fabric](https://fabric.io) (its successor) does. I'll need to have an internal discussion about switching to Fabric. In the interim, could you remove this and rip out our usage of Crashlytics?

lfaraone:
This PR combines a reformat of the entire codebase with the switch to Gradle. It'd be a lot easier to review if those were separate commits, and a lot more useful to future `git annotate`rs.

app/build.gradle
se-bastiaan: Crashlytics is a part of Fabric now. If you were using Crashlytics before then you're using Fabric now, if you want it or not. You can have your internal discussion about switching to Fabric, but it would be more about 'should we use Fabric or not use Fabric?' since you basically have not other option.

Issue #1, opened 2015-09-26
Update login form to support zulip servers that aren't zulip.com


timabbott:
The most important feature missing from the Android app right now is that it currently hardcodes talking to zulip.com.  It should be a pretty straightforward change to have the user type in their server on the login page.

willdeed:
+1 to this. I'm not sure it's that straight forward as there may be UI changes and other use cases to consider when dealing with a connection to a server that's not controlled by Zulip, but it's definitely an important feature from my perspective as well. 

I wish I could help with the coding, but I'm lacking the Android dev skillz.

scar45:
I was tinkering with it tonight, and I was able to get the app building via Gradle in Android Studio after fixing Play Services and commenting out Crashlytics.

I was then able to change the URL that the Android app uses to authenticate against in **ZulipApp.java @ L:126**:

```java
    public String getServerURI() {
//        if (getEmail().equals("iago@zulip.com")) {
//            return "http://10.0.2.2:9991/api/";
//        }
        return "https://MY.SERVER.HERE/";
    }
```

...and when debugging, I see it hit that connection, then (presumably) after my login is validated, it POSTs "v1/fetch_api_key" which returns a 404 error. I dug around the server source and found what I think is the correct API endpoint. Here's what I changed it to in **AsyncLogin.java @ L:31**:

```java
    public final void execute() {
        execute("POST", "api/v1/fetch_api_key");
    }
```

...but it still doesn't appear to work for me. I'm not sure, but I think that URL is looking for the username + API key, instead of the username + password which I think it's being fed. Just shooting in the dark at this point though.

I am not an Android dev by any means, but after getting the server-side working great, I am dying to get the app working! Hope this can help in some way.

leoluk:
Needs a custom Google API key too: https://github.com/zulip/zulip-android/blob/3f58f50112dc6ec1683570f90ee56a78805809bf/src/com/zulip/android/Notifications.java#L21

scar45:
Nice find @leoluk -- I will look into registering a Google API Key. I do have a question about the counterpart in zserver though. I setup Zulip on my own VM using the production tarball. It appears that after editing /etc/zulip/settings.py, the file contents get copied to /home/zulip/deployments/current/zproject/local_settings.py, which is the file above that you referenced. 

[...]

PR #16, opened 2015-11-02
Fix SSL/TLS hostname-not-valid error

ff-:
As I detailed in #1 I rewrote the `HTTPRequest` class using the native `java.net.HttpsURLConnection` because the Apache Http lib doesn't support [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) and because starting from Android 5.1 all the `org.apache.http` classes are being [deprecated](https://developer.android.com/about/versions/android-5.1.html#http).

Not supporting SNI means that when your Zulip instance is hosted on a virtualhost the TLS auth fails even if the certificate is valid and trusted.

### DIFF
--- a/src/com/zulip/android/HTTPRequest.java
+++ b/src/com/zulip/android/HTTPRequest.java
@@ -1,137 +1,150 @@
 package com.zulip.android;
 
-import java.io.ByteArrayOutputStream;
+import java.io.BufferedInputStream;
+import java.io.BufferedReader;
+import java.io.DataOutputStream;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLEncoder;
+import java.util.HashMap;
+import java.util.Map;
 
-import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
-import org.apache.http.NameValuePair;
-import org.apache.http.StatusLine;
 import org.apache.http.client.HttpResponseException;
-import org.apache.http.client.entity.UrlEncodedFormEntity;
-import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
-import org.apache.http.client.methods.HttpGet;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.client.methods.HttpPut;
-import org.apache.http.client.methods.HttpRequestBase;
-import org.apache.http.client.utils.URLEncodedUtils;
-import org.apache.http.message.BasicNameValuePair;
-import org.apache.http.params.HttpConnectionParams;
 
-import android.net.http.AndroidHttpClient;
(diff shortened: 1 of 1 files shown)


smarx:

@ff-, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/) and update the thread so we can consider merging your code.

ff-:
I signed the CLA.

src/com/zulip/android/HTTPRequest.java
```diff
@@ -1,137 +1,150 @@
...
+                throw new IOException("Wrong HTTP method.");
+            }
+
+            if (method.equals("GET")) {
+                url = new URL(app.getServerURI() + path + "?" + getPostDataString(this.postDataParams));
             }
 
+            conn = (HttpsURLConnection) url.openConnection();
+            conn.setRequestMethod(method);
+            conn.setRequestProperty("Authorization", "Basic "
```
lfaraone: Can we switch to [`java.net.Authenticator`][1] instead here?

[1]: https://developer.android.com/reference/java/net/Authenticator.html

src/com/zulip/android/HTTPRequest.java
```diff
@@ -1,137 +1,150 @@
...
-                            + Base64.encodeToString(authstr.getBytes(),
-                                    Base64.NO_WRAP));
             // timeout after 60 seconds in ms
-            HttpConnectionParams.setSoTimeout(request.getParams(), 60 * 1000);
+            conn.setConnectTimeout(60 * 1000);
+            conn.connect();
+
+            InputStream responseStream;
+
+            if(conn.getResponseCode() >= HttpStatus.SC_BAD_REQUEST)
```
lfaraone: We should move off `HttpStatus` if possible, since leaving it in requires us to continue using deprecated libs.

src/com/zulip/android/HTTPRequest.java
```diff
@@ -1,137 +1,150 @@
...
+            while ((line = responseStreamReader.readLine()) != null) {
+                stringBuilder.append(line).append('\n');
+            }
+
+            responseStreamReader.close();
+            responseString = stringBuilder.toString();
+            responseStream.close();
+
+            if (conn.getResponseCode() != HttpStatus.SC_OK) {
+                Log.e("HTTP", conn.getResponseMessage());
```
lfaraone: We're logging this data twice, once here and once on 147.

src/com/zulip/android/HTTPRequest.java
```diff
@@ -1,137 +1,150 @@
...
-        }
+        return result.toString();
     }
 
     String execute(String method, String path) throws IOException {
-        AndroidHttpClient httpclient = AndroidHttpClient.newInstance(app
-                .getUserAgent());
-
-        String responseString = null;
+        String responseString = "<empty>";
```
lfaraone: Why is this changed here from `null`?

niftynei:
@ff-  Since we moved over to OKHttp, this patch is largely obsolete.  If it's still a problem, please open an issue. 

Thanks for your work on this.

PR #17, opened 2015-11-03
Fix #15: integration bots profile images fetch

### LINKED ISSUE #15: Integration bots profile image make the app crash
When there's a bot in the stream, the app tries to fetch its profile image from Gravatar, while its url is a relative path for the server.

ff-:
Bot integrations profile images paths are relative to the server root, with a path like `/user_avatars/<BOT_IMAGE_ID>` instead of `https://secure.gravatar.com/users/<USERHASH>` which is typical for real users.

The app tried to fetch them without checks, failing for them not being correct absolute URLs.

### DIFF
--- a/src/com/zulip/android/GravatarAsyncFetchTask.java
+++ b/src/com/zulip/android/GravatarAsyncFetchTask.java
@@ -1,7 +1,10 @@
 package com.zulip.android;
 
+import java.io.BufferedInputStream;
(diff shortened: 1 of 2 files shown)


smarx:

@ff-, it looks like you've already signed the Dropbox CLA. Thanks!

niftynei:
fixed by #211

PR #20, opened 2015-11-28
Add an server option to the login screen

armooo:
Allow users to set a custom zulip server. We special case the URL format
of api.zulip.com to not include /api/.

### DIFF
--- a/res/layout/login.xml
+++ b/res/layout/login.xml
@@ -38,13 +38,23 @@
         android:hint="@string/password"
         android:inputType="textPassword" />
 
+    <EditText
+        android:id="@+id/server_url"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_alignParentLeft="true"
+        android:layout_below="@+id/password"
+        android:ems="10"
+        android:hint="@string/server_url"
+        android:inputType="textUri" />
+
     <Button
         android:id="@+id/login"
         style="android:buttonStyle"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_alignParentLeft="true"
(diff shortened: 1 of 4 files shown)


timabbott:
Awesome, thanks for doing this @armooo!!!   I guess you should tag this as fixing #1 :).  To what extent have you tested this?  @lfaraone can you review?

armooo:
I have tested by this logging into my own zulip install. I don't have access to a zulip.com realm that supports password access so that flow is untested. I have a debug build that can be used to test on another realm. https://www.dropbox.com/s/ujk2usxpzro5qgx/zulip-debug.apk?dl=0

I have never been able to test the OAuth2 login flow on a local build. I think it requires the release signing key to work correctly. I also suspect this will not work with self hosted installs due to missing secrets.

timabbott:
OK that seems pretty reasonable given the nature of the change; it'd be great to get the OAuth2 flow working anywhere other than zulip.com but I don't see a reason to block this PR on that...

chauffer:
Seems nice, is this gonna merge?

christian667:
Looks good, we need this as well, please merge and release soon :)

SeMeKh:
Any progress on this one? We're pretty eager to see this PR land in `master`.

magro:
+1

bpereto:
+1

willdeed:
+22

GRbit:
+1

timabbott:
@lfaraone I seem to recall you looked at this briefly and determined some changes were needed; but I don't remember the details, can you post them here?  Then one of the folks who are actively working on the project can take a look at finishing this...

niftynei:
merged as a part of #52

PR #21, opened 2016-01-04
Feature/gradle.android.studio.2.x

kosiara:
Changes:

- removed *.jar dependencies
- support for Android Studio 2.x
- added gradle build system
- changed project structure
- fixed: nullpointer when there was no avatar in user's profile


lfaraone:
Since we've migrated the app to Gradle, some of the changes here aren't needed anymore. Could you submit a new PR with those broken out if they still apply?

Thanks!

Issue #34, opened 2016-03-25
Allow narrowing to stream / thread 


niftynei:
On the web interface, clicking on the stream or thread will narrow the view to just those messages.  The Android app should do the same.

bgilbert:
It does: just long-press on the message and select the appropriate narrow command.

niftynei:
Oh cool. That's a hard to discover feature. It's good to know it exists.

It's worth pointing out that that doesn't address the issue highlighted in this ticket -- that the Android app behavior is different than the web client, which is what this issue proposes fixing.

niftynei:
Hm. I realized that this intent wasn't explicit in the message title, owing to the fact that I was under the mistaken assumption the functionality didn't exist. Thanks for pointing it out!

PR #35, opened 2016-03-26
Update UI and Improve Usability

niftynei:
A few small changes to the UI and adds some much needed navigation to messages in threads.  Includes updating to material design, so that the compose dialogs look more modern.  Also adds debug icons so that it's easy to tell the difference between a 'debug' build and a prod build, if you have them both loaded on your phone.

Incorporates changes requested in #34.

screenshot_20160326-105607
screenshot_20160326-105637
screenshot_20160326-105644
screenshot_20160326-105653
screenshot_20160326-105705
screenshot_20160326-105824
screenshot_20160326-105905


smarx:

timabbott:
The mockups look pretty good!  

A few things I noticed on looking at this (not reviewing the code carefully since I've done very little Android development):
* Looks like there's a small merge conflict with the .gitignore file from your other branch.
* In 6c4afce1a4a6a7539f9e5f653845dd9065e92126, why does it make sense to reverse the if statement in the R.id.search?  Probably makes sense to just put the reason in the commit message and refer back here.
* A similar note is that the "You and" thing I bet still looks funny if you PM yourself.
* One thought that I don't think needs to be resolved here but I was reminded of by the mocks is that it's somewhat problematic that there's no way to see the message you're replying to in the compose box.  Might be worth opening an issue for trying to fix that.

@lfaraone can you take a look?

niftynei:
- the merge conflict with `.gitignore` has been resolved
- I added a longer explanatory note to the commit message; the short version is that for honeycomb+ two edit text boxes were being activated when you clicked search (a dialog & a field in the actionbar). only the actionbar edit text should show up.
- The "You and " PM _was_ a message to myself :) 
- Just opened issue #43 for this :+1:

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -418,6 +418,7 @@ public void doNarrow(NarrowFilter filter) {
 
     @Override
     public void onNarrow(NarrowFilter narrowFilter) {
+        // TODO: check if already narrowed to this particular stream/subject
```
lfaraone: squash this into cbf90468

niftynei: is there a reason for squashing this?  it's a bit more general than just to that particular commit in that it applies to the current code AS-IS as well.

niftynei:
@lfaraone bump. is there anything else I can change on this so we can get it merged in?

Issue #10, opened 2015-10-19
Add a way to mute a topic


bgilbert:
There doesn't seem to be any UI for muting a topic.

bgilbert:
This could go in the context menu available by long-pressing a message.

Swati4star:
@kunall17 This enhancement is there in the app. Shouldn't we close this one now?

kunall17:
@Swati4star Nope, its not there the code is there, I had implemented this working but there was no sync with the server at the time so it was temporarily removed If you can work on the server for the API it would be great! 

@timabbott Do we have endpoint API for muting topic/stream/private message?

timabbott:
Hmm, we don't have an API endpoint for muting right now, but it'd be easy to add one; we'd basically just need to move `set_muted_topics` from a legacy non-API URL to an API URL.

timabbott:
(I can help with this if someone is interested in using that feature)

Issue #9, opened 2015-10-19
Muted topics are not hidden


bgilbert:
Incoming messages on muted topics are not hidden, either in the home view or in views of individual streams.

timabbott:
Yep muted topics are not supported in the Android app.

ryansnowboarder:
I'm trying to fix some things on the Android app and I'm hitting a wall while logging in. How do I do that (requires a @zulip.com login)?

niftynei:
Fixed as of #42.

PR #42, opened 2016-03-31
Implement Mute topics 

kunall17:
#9,#10
Muting topics in the HomeView removes them from the homeView, and are visible in their respective streams.
Topics can be unmuted from the context menu of their streams.

Screenshots:
**Muting Topics by context menu in the Homeview**

**Unmuting Topics by context menu in the Streams**


smarx:

timabbott:
Just to clarify, in Zulip we have two concepts, muted streams and muted topics, and the way they currently work is that messages muted either way don't show up on the home view, while messages to muted topics shouldn't show up either in the home view or in the stream view (in either caes, unless the user is mentioned).  `static/js/muting.js` is a helpful reference for topic muting, while grepping the server/web codebase for `in_home_view` will find the code for stream muting, since that is what it is called in the code.

kunall17:
@timabbot but if a user want's to see the muted messages even if they are not tagged then where will they see?
In the web there is listing of the topics in a stream where muted topics are displayed as well and if we unmute them we can view them, But in android we don't have that currently!

Possible solution is we could list the topics in a stream in the sidebar like web?

timabbott:
Yeah, that's a possible solution; if we can't make that work from a UI perspective, then perhaps your approach may be a reasonable hack for now, but I think it's won't solve the full problem, and if we do that, we should make it clear in code comments how the model differs from how muting works on the web.

timabbott:
(One of our Android experts should review this branch, I just wanted to mention the details on how muting works on web since it's not well-documented that this is the model)

kunall17:
@timabbott so should i start working on this solution?

timabbott:
Yeah, I think so.  My current thinking is the right approach is to mimic the model on web (with muted streams only visible when narrowed to the stream, and muted topics only visible when narrowed to the topic), but adding a way to see the list of topics on a stream.

@niftynei, do you have ideas for how to display the topics within a stream?  One idea would be to display it in the left drawer, with the topics for the stream one is currently narrowed to indented a bit.

(As a sidenote, we should probably add some sort of highlighting of the currently narrowed stream in the left sidebar if you're narrowed, since it's pretty lame that we don't do that).

app/src/main/java/com/zulip/android/MessageListFragment.java
```diff
@@ -53,6 +54,9 @@
 
         void openCompose(final MessageType type, String stream, String topic,
                          String pmRecipients);
+        void addToList(Message message);
+        void unMuteTopic(int StreamID,String subject);
+        void MuteTopic(int StreamID,String subject);
```
niftynei: please use standard Java conventions for function names here. 

> Compose method names using mixed case letters, beginning with a lower case letter and starting each subsequent word with an upper case letter.

source: https://www.cwu.edu/~gellenbe/javastyle/method.html

app/src/main/java/com/zulip/android/MessageListFragment.java
```diff
@@ -93,16 +98,17 @@ public MessageListFragment() {
...
-
+    List<Message> messageList;
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         if (getArguments() != null) {
             filter = getArguments().getParcelable(PARAM_FILTER);
         }
-
+        muted_messages = new ArrayList<>();
```
niftynei: Please use conventional Java variable names here. ie `mutedMessages` instead of `muted_messages`.

app/src/main/java/com/zulip/android/MessageListFragment.java
```diff
@@ -275,6 +281,13 @@ public void onCreateContextMenu(ContextMenu menu, View v,
         if (msg.getType().equals(MessageType.STREAM_MESSAGE)) {
             MenuInflater inflater = getActivity().getMenuInflater();
             inflater.inflate(R.menu.context_stream, menu);
+            if ((app.isTopicMute(msg.getStream().getId(), msg.getSubject()))) {
+                //Muted Topic
+                menu.findItem(R.id.mute_topic).setTitle("Unmute Topic");
```
niftynei: Strings belong in `strings.xml`

app/src/main/res/menu/context_stream.xml
```diff
@@ -4,6 +4,7 @@
     <item android:id="@+id/reply_to_sender" android:title="Reply to Sender"></item>
     <item android:id="@+id/narrow_to_stream" android:title="Narrow to this stream"></item>
     <item android:id="@+id/narrow_to_subject" android:title="Narrow to this topic"></item>
+    <item android:id="@+id/mute_topic" android:title="Mute Topic"></item>
```
niftynei: User visible strings should be put in `strings.xml`.

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
             unreadMessageHandler.sendEmptyMessageDelayed(0, 2000);
         }
     }
+    public void MuteTopic(int streamId, String subject) {
```
niftynei: See above comment about using proper Java naming conventions for method names.

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
...
+        System.out.println((streamId + "-" + subject));
```
niftynei: Can you remove this log line so it won't get merged into master?

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
...
+        SharedPreferences.Editor editor = settings.edit();
+        editor.putStringSet("muted_topics", muted_topics);
+        editor.commit();        
+    }
+
+    public void UnMuteTopic(int streamId, String subject) {
+        muted_topics.remove(streamId + "-" + subject);
+        SharedPreferences.Editor editor = settings.edit();
+        editor.putStringSet("muted_topics", muted_topics);
+        editor.commit();
```
niftynei: `commit` is fine, but it's generally best practice to use the async command `apply`.  See http://stackoverflow.com/a/5960732

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
...
+    }
+
+    public boolean isTopicMute(int streamId, String subject) {
+        if (muted_topics.contains(streamId + "-" + subject))
+            return true;
+        return false;
+    }
+    public boolean isTopicMute(String idAndsubject) {
```
niftynei: Capitalization on `idAndsubject` should be `idAndSubject`.

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
...
+        if (muted_topics.contains(idAndsubject))
+            return true;
```
niftynei: the `if` here is superfluous. you can simply return ie:

    return muted_topics.contains(idAndSubject)

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
...
+    public void MuteTopic(int streamId, String subject) {
+        System.out.println((streamId + "-" + subject));
+        muted_topics.add(streamId + "-" + subject);
+        SharedPreferences.Editor editor = settings.edit();
+        editor.putStringSet("muted_topics", muted_topics);
+        editor.commit();        
+    }
+
+    public void UnMuteTopic(int streamId, String subject) {
+        muted_topics.remove(streamId + "-" + subject);
```
niftynei: It seems like adding a "-" is a generic operation that you're doing every time you put/fetch a streamID & subject to the Shared Preferences.  It's a good idea to break this out into a function and use the same logic every place that this operation is required.

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
...
+        SharedPreferences.Editor editor = settings.edit();
+        editor.putStringSet("muted_topics", muted_topics);
```
niftynei: `putStringSet` is buggy when used in this manner.  See http://developer.android.com/reference/android/content/SharedPreferences.html#getStringSet%28java.lang.String,%20java.util.Set%3Cjava.lang.String%3E%29.  You'll need to either create a new object to pass in to save, or clear out the value before saving.

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +264,29 @@ public void markMessageAsRead(Message message) {
...
+        editor.commit();
+    }
+
+    public boolean isTopicMute(int streamId, String subject) {
+        if (muted_topics.contains(streamId + "-" + subject))
```
niftynei: the `if` here is superfluous. you can simply return ie:

    return muted_topics.contains(streamId + "-" + subject)

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -54,6 +57,48 @@
...
+                public int compare(Message message1, Message message2) {
+                    return message1.getTimestamp().compareTo(message2.getTimestamp());
+                }
+            });
+            homeList.adapter.notifyDataSetChanged();
+        }
+    }
+
+    @Override
+    public void MuteTopic(int StreamID, String subject) {
```
niftynei: Per Java parameter naming conventions, `StreamID` should be `streamId`.

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -54,6 +57,48 @@
         MessageListFragment.Listener {
 
     ZulipApp app;
+    List<Message> muted_topics;
```
niftynei: `muted_topics` should be `mMutedTopics` or `mutedTopics`.

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -54,6 +57,48 @@
...
+    @Override
+    public void addToList(Message message) {
+        muted_topics.add(message);
+    }
+
+    @Override
+    public void unMuteTopic(int StreamID, String subject) {
```
niftynei: Per Java conventions, `StreamID` should be renamed `streamId`.

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -54,6 +57,48 @@
...
+        app.UnMuteTopic(StreamID, subject);
+        int added = 0;
+        for (Iterator<Message> messageIterator = muted_topics.iterator(); messageIterator.hasNext(); ) {
```
niftynei: A cleaner way to do this would be by implementing `equals` and `hashCode` on the `Message` object; this would enable you to be able to call `muted_topics.remove(message)`.

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -54,6 +57,48 @@
...
+            Message message = messageIterator.next();
+            if (message.getStream().getId() == StreamID && message.getSubject().equals(subject)) {
+                messageIterator.remove();
+                homeList.adapter.add(message);
+                added++;
+            }
+        }
+        if (added > 0) {
+            Collections.sort(homeList.messageList, new Comparator<Message>() {
```
niftynei: This might be nicer to have `Message` implement the Comparable interface and move the logic into the `Message` class.

app/src/main/java/com/zulip/android/MessageListFragment.java
```diff
@@ -315,6 +328,13 @@ public boolean onContextItemSelected(MenuItem item) {
             case R.id.copy_message:
                 copyMessage(message);
                 return true;
+            case R.id.mute_topic:
+                if (app.isTopicMute(message.getStream().getId(), message.getSubject())) { //Unmute topics
```
niftynei: A cleaner API might be to have `isTopicMute` accept a `Message` object.

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -181,6 +226,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
         setContentView(R.layout.main);
 
+muted_topics = new ArrayList<>();
```
niftynei: Spacing here seems off. Can you indent appropriately?

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -54,6 +57,48 @@
...
+                }
+            });
+            homeList.adapter.notifyDataSetChanged();
+        }
+    }
+
+    @Override
+    public void MuteTopic(int StreamID, String subject) {
+        app.MuteTopic(StreamID, subject);
+        for (int i = homeList.adapter.getCount() - 1; i >= 0; i--) {
```
niftynei: depending on how large the `homeList`, this might be a bit of an expensive operation. it might be worthwhile to add some type of indexed look up for message topics to the `homeList` adapter.  not strictly necessary for this PR tho :)

niftynei:
Showing the topics in the drawer (same as web) seems like a great way to fix the UI issue, but that's a larger task than this PR.  My inclination is that it's ok in the interim to force users to use the web UI to unmute topics, with the idea that mobile apps are more streamlined (so it's ok to see less information in the short-period while we work out the best way to display it).

Alternatively, I'd propose adding a separate activity/screen to view/edit all muted topics & streams.  You could add a link to it in the home activity's overflow menu.

niftynei:
Also, I don't really know much about how muting works on Zulip, but @kunall17 I don't see anywhere that you're pre-populating the list of muted topics/streams.  Is this something that you'd need to fetch from the server?  As is it seems that your muted topics/streams would only be app-local, which has a lot of potential for confusion for users that use both the Web and app clients.

niftynei:
also @kunall17 you might take a gander at the AOSP [code style guidelines](https://source.android.com/source/code-style.html) and the Google [Java styles](https://google.github.io/styleguide/javaguide.html).  I might not have marked all of them, but there's a bunch of style problems in this PR.

kunall17:
Hello @niftynei, I have updated the code style!
And in the commit (Fetch Muted Topics from the server) I have added code to parse the JSON and get the Muted Topics from the server!

app/src/main/java/com/zulip/android/MessageListFragment.java
```diff
@@ -275,6 +281,13 @@ public void onCreateContextMenu(ContextMenu menu, View v,
         if (msg.getType().equals(MessageType.STREAM_MESSAGE)) {
             MenuInflater inflater = getActivity().getMenuInflater();
             inflater.inflate(R.menu.context_stream, menu);
+            if ((app.isTopicMute(msg))) {
+                //Muted Topic
+                menu.findItem(R.id.mute_topic).setTitle(getResources().getString(R.string.un_mute_topic));
+            } else {
+                if (filter != null)
```
niftynei: in what condition will filter not be null?

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +285,21 @@ public void markMessageAsRead(Message message) {
...
+    }
+
+    public void unMuteTopic(Message message) {
+        mutedTopics.remove(message.getStream().getId() + message.getSubject());
+        SharedPreferences.Editor editor = settings.edit();
+        editor.putStringSet(MUTED_TOPIC_KEY, new HashSet<String>(mutedTopics));
+        editor.apply();
+    }
+
+    public boolean isTopicMute(Message message) {
```
niftynei: this API looks much cleaner 👍

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -261,4 +285,21 @@ public void markMessageAsRead(Message message) {
             unreadMessageHandler.sendEmptyMessageDelayed(0, 2000);
         }
     }
+    public void muteTopic(Message message) {
+        mutedTopics.add(message.getStream().getId() + message.getSubject());
```
niftynei: I'd still prefer to have the operation for generating the 'key' for a muted topic to be a separate function, or even embed it into the `Message` object. for example:

    mutedTopics.add(message.concatStreamAndTopic());
    ...
    // in Message.java
    private String concatStreamAndTopic() {
        return getStream().getId() + getSubject();
    }

niftynei: Also might need to check for `null` on the `getStream` call.

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -146,6 +153,23 @@ public String getUserAgent() {
...
+    public void addToMutedTopics(JSONArray jsonArray) {
+        Stream stream;
+
+        for (int i = 0; i < jsonArray.length(); i++) {
+            try {
+                JSONArray mutedTopic = jsonArray.getJSONArray(i);
+                stream = Stream.getByName(this, mutedTopic.get(0).toString());
+                mutedTopics.add(stream.getId() + mutedTopic.get(1).toString());
+            } catch (JSONException e) {
+                e.printStackTrace();
```
niftynei: Please log to Log.e instead of printing the stacktrace.

    Log.e(TAG, e);

niftynei:
@kunall17 nice work! this looks much better 💯 

Is there a way to post "Mute topic" actions back up to the server?  If not, we shouldn't present the user with a button to mute/unmute anything until we can sync state back to the server.

app/src/main/java/com/zulip/android/ZulipApp.java
kunall17: @niftynei 
I think the stream won't be null as these are only for the muting topics are only for the stream messages!

kunall17:
@niftynei 
Updated the Code.

And I'm not sure if there is a way to post Mute topic action on the server.

niftynei:
@kunall17 if that's the case, then we should remove the button allowing the mute/unmute action.

kunall17:
@timabbott can you help please?
I want to send a POST request to the server for muting/unmuting a topic, how can i do this?

I tried to find out I got this [file](https://github.com/zulip/zulip/blob/2adf6d822ff1799cbba81f6288c7aedf188a1af7/zproject/urls.py#L135) which shows mute topics for the webClient

timabbott:
There's no current API route available for changing topic muting; see
https://github.com/zulip/zulip/issues/611 for what needs to be done to change that.  

The mechanism for stream muting involves setting the `in_home_view` flag on a subscription object, so that should be possible with the current API via `PATCH users/me/subscriptions`

kunall17:
@timabbott ok thankyou, Will look into it..

kunall17:
@timabbott bump, please..

timabbott:
Ahh OK @kunall17, I hadn't noticed you'd edited your previous message.

I defer to @niftynei on how to manage this PR.  I think ultimately we probably want to both display streams/topics correctly on Android as well as control the muting status on Android, but often it can be a good idea to try to finish the first part really well (and get it merged) as a separate project from the second part.

kunall17:
Now I have, 
- [x] Removed muting/unmuting topics from the context menu and muted topics are hidden from homeView.
- [x] Change TextColor of the muted streams in the Stream Drawer.

I am still using the SharedPreferences for saving the muted topics, as in when API routes are developed, I will add unmute/mute topics and streams so there it will be useful.

niftynei:
@kunall17  reviewing this, i'm seeing muted topic messages in the stream.  this seems broken. can you explain what behaviour I should expect for muted topics? 

I'll leave a comment in line as well, but also worth noting that the text-color changing you've implemented for streams in the drawer is buggy -- it's greying out streams other than the ones I've muted.

app/src/main/java/com/zulip/android/ZulipActivity.java
```diff
@@ -88,6 +112,9 @@ public boolean setViewValue(View arg0, Cursor arg1, int arg2) {
                 case R.id.name:
                     TextView name = (TextView) arg0;
                     name.setText(arg1.getString(arg2));
+                    //Change color in the drawer if this stream is inHomeView only.
+                    if (!Stream.getByName(app, arg1.getString(arg2)).getInHomeView())
+                        name.setTextColor(ContextCompat.getColor(ZulipActivity.this, android.R.color.tertiary_text_light));
```
niftynei: When these views get recycled, the text stays grey (so streams that aren't mute have grey labels 😞 ).  One way to fix this is to reset the text back to the default color if the Stream isInHomeView().

kunall17: Ohh yeah sorry, didn't notice this one as I muted all the streams in my server.!

kunall17:
@niftynei in these new commits I have added a icon of muted topics to the header of the message tile.

niftynei:
looks good to me. thanks for all the work on this @kunall17! 👍

kunall17:
@niftynei thankyou for merging this one. :)

PR #63, opened 2016-05-09
fix NPE bug and clean up logic for setting the title bar

niftynei:
Samsung user was experiencing an NPE crash on startup.  This patch fixes the NPE error (getActionBar() was returning null) and condenses the logic for this into one place.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -460,25 +460,22 @@ public class ZulipActivity extends FragmentActivity implements
         currentList = list;
 
         NarrowFilter filter = list.filter;
-
         if (filter == null) {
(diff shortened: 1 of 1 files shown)


smarx:

niftynei:
@kunall17 would you mind reviewing this?  `getActionBar()` was causing an NPE for some older Samsung phones (/me shakes fist at Samsung).

kunall17:
Looks good, but i never got this error in a Sony Phone, do you know know why is this causing in a samsung?

niftynei:
The problem appeared on an older Samsung Galaxy S4.  Samsung is notorious for re-writing the Android standard libraries, giving them bugs that don't appear on other platforms. I don't know exactly why that version of Samsung crashes, but if I had to guess it'd be that they call the method before the view has been instantiated.

niftynei:
Thanks @kunall17 !

PR #68, opened 2016-06-16
Bugfix api calls and gcm key

markusthegeek:
Needed to add slashes to all api calls as in "/v1/...". Also fix the problem with the empty current_key value that is now needed with the latest GMS service.

### DIFF
--- a/app/src/debug/google-services.json
+++ b/app/src/debug/google-services.json
@@ -29,7 +29,9 @@
           "client_type": 3
         }
       ],
-      "api_key": [],
+      "api_key": [{
+          "current_key": ""
+        }],
       "services": {
         "analytics_service": {
           "status": 1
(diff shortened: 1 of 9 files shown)


smarx:

@markusthegeek, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

smarx:
				
@markusthegeek, thanks for signing the CLA!

niftynei:
hey @markusthegeek thanks for the PR. What version of GMS does this problem crop up on?

niftynei:
The root cause of this issue was a missing backslash in the parsing code for a user-provided domain. Issue is fixed in #78 .

I believe that #70 allow us to remove the `google-services.json`, rendering the empty API key issue moot.  If this isn't the case, we can re-open this suggested fix.

Thanks!

Issue #48, opened 2016-04-12
Build a slick login experience when DevAuthBackend is enabled on web


timabbott:
This is probably actually a joint Android/server project, but it'd be great if when you pointed the Android app at a Zulip server that's using DevAuthBackend (which shows the grid of users to login as on the web for development), if we could provide a similar experience on mobile where it just shows a list of users and you can click to login as any of them, so one doesn't need to spend time dealing with passwords when developing against a development server.

ryansnowboarder:
I'll work on this after I finish profiling the app startup, which should be sometime today or tomorrow.

niftynei:
This has been merged! :)

PR #45, opened 2016-04-08
Implemented chatBox to directly reply. 

kunall17:
Features covered in this PR (#8)- 

- [x] Send message without dialog box
- [x] Tap on the chat message textView to select the subject or person in the chatBox and send message
- [x] Tap on the header tile to narrow and select the subject or person in the chatBox
- [x] Switch views for the Stream and Private.

ezgif-4210950663

Issue #29, opened 2016-03-20
Best Network Library suited for Android Project


kunall17:
As I and @timabbott had a conversation for the best network library suited for the project I did some test's for 4 libraries which are HttpURLConnection, Retrofit, Volley and the current Apache Client.

I did these test on a 1.5 Mbps internet connection to fetch from the server having custom headers for authorization and record the time taken in millisecond. Total size to fetch was **487 Kb** 

So I conducted 5 test's with each of the libraries and got the following results- 

|   Name   | HTTPURLConnection | RetroFit | Apache Client | Volley |
|:--------:|:-----------------:|:--------:|:-------------:|:------:|
|          |                   |          |               |        |
| 1st Test |        3172       |   2886   |      3255     |  3127  |
| 2nd Test |        3157       |   2776   |      3275     |  3055  |
| 3rd Test |        3121       |   2835   |      3292     |  3082  |
| 4th Test |        3195       |   2859   |      3223     |  3145  |
| 5th Test |        3175       |   2801   |      3283     |  3048  |

Therefore I found the average values - 

|        Name       | Average Time Taken (ms) | Minimal Lines of Code |
|:-----------------:|:-----------------------:|:---------------------:|
| HTTPURLConnection |           3164          |           28          |
|      Retrofit     |           2831          |           9           |
|   Apache Client   |           3265          |           24          |
|       Volley      |           3091          |           25          |

I have also uploaded the [project test's](https://github.com/kunall17/zulip-network-tests) I built for these libaries
Now by seeing the result we observe Retrofit is the best in these libraries in terms of speed, total code.
What are your suggestions for replacing the current library?

timabbott:
Cool, yeah, that seems like reasonably compelling evidence that Retrofit is faster (at least on your test network) and less code.

ryansnowboarder:
Hmm... interesting! 487 Kb sounds reasonable for what that app would normally fetch upon startup. The REAL question: is that time difference actually a significant contribution to speeding up the app startup? I'm sure there's some other things going on that might slow things down...

ryansnowboarder:
Would you happen to have a speed test screen capture for that internet connection? I'm curious as to what the latency of the connection was - not that it's a likely issue we're all being sent through a proxy on a satellite, but might as well cover all the bases.

kunall17:
[...]

PR #69, opened 2016-06-19
Replaced Old Apache HTTP Libraries with OkHttpClient

kunall17:
Espresso Test's are already included in #45 and #48!

### DIFF
--- a/app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
+++ b/app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
@@ -8,7 +8,6 @@ import java.util.HashMap;
 import java.util.concurrent.Callable;
 
 import org.apache.commons.lang.time.StopWatch;
-import org.apache.http.client.HttpResponseException;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -26,10 +25,12 @@ import com.zulip.android.util.ZLog;
 import com.zulip.android.activities.ZulipActivity;
 import com.zulip.android.ZulipApp;
 
+import okhttp3.Response;
+
 public class AsyncGetEvents extends Thread {
     private static final String TAG = "AsyncGetEvents";
-    public static final String ASYNC_GET_EVENTS = "asyncGetEvents";
-    public static final String POINTER = "pointer";
+    private static final String ASYNC_GET_EVENTS = "asyncGetEvents";
+    private static final String POINTER = "pointer";
     ZulipActivity activity;
     ZulipApp app;
 
@@ -77,7 +78,8 @@ public class AsyncGetEvents extends Thread {
 
         StopWatch watch = new StopWatch();
         watch.start();
-        String responseData = request.execute("POST", "v1/register");
+        request.setMethodAndUrl("POST", "v1/register");
+        String responseData = request.execute().body().string();
         watch.stop();
         Log.i("perf", "net: v1/register: " + watch.toString());
 
@@ -108,8 +110,24 @@ public class AsyncGetEvents extends Thread {
                     if (!registeredOrGotEventsThisRun) {
                         request.setProperty("dont_block", "true");
                     }
-                    JSONObject response = new JSONObject(request.execute("GET",
-                            "v1/events"));
+                    request.setMethodAndUrl("GET", "v1/events");
+                    Response httpResponse = request.execute();
+
+                    String json = httpResponse.body().string();
+                    if (!httpResponse.isSuccessful()) {
+                        String msg = httpResponse.message();
+                        if (msg.contains("Bad event queue id")
+                                || msg.contains("too old")) {
+                            // Queue dead. Register again.
+                            Log.w(ASYNC_GET_EVENTS, "Queue dead");
+                            app.setEventQueueId(null);
+                        }
+                        Log.i("WRONG", "run: " + json);
+                        continue;
+                    } else {
+                        Log.i("OkHttp200GE", json);
+                    }
+                    JSONObject response = new JSONObject(json);
 
                     JSONArray events = response.getJSONArray("events");
                     if (events.length() > 0) {
@@ -131,20 +149,9 @@ public class AsyncGetEvents extends Thread {
                             }
                         });
                     }
-                } catch (HttpResponseException e) {
-                    if (e.getStatusCode() == 400) {
-                        String msg = e.getMessage();
-                        if (msg.contains("Bad event queue id")
-                                || msg.contains("too old")) {
-                            // Queue dead. Register again.
-                            Log.w(ASYNC_GET_EVENTS, "Queue dead");
-                            app.setEventQueueId(null);
-                            continue;
-                        }
-                    }
-                    backoff(e);
                 } catch (SocketTimeoutException e) {
                     Log.e(TAG, e.getMessage(), e);
+                    ZLog.logException(e);
                     // Retry without backoff, since it's already been a while
                 } catch (IOException e) {
                     if (request.aborting) {
@@ -156,6 +163,7 @@ public class AsyncGetEvents extends Thread {
                 } catch (JSONException e) {
                     backoff(e);
                 }
+                Thread.sleep(1000);
             }
         } catch (Exception e) {
             ZLog.logException(e);
--- a/app/src/main/java/com/zulip/android/networking/AsyncLogin.java
+++ b/app/src/main/java/com/zulip/android/networking/AsyncLogin.java
@@ -1,7 +1,5 @@
 package com.zulip.android.networking;
 
-import org.apache.http.HttpStatus;
-import org.apache.http.client.HttpResponseException;
 import org.json.JSONException;
 import org.json.JSONObject;
 
@@ -15,12 +13,13 @@ import com.zulip.android.util.ZLog;
 import com.zulip.android.ZulipApp;
 
 public class AsyncLogin extends ZulipAsyncPushTask {
-    public static final String UNREGISTERED = "unregistered";
-    public static final String DISABLED = "disabled";
+    private static final String UNREGISTERED = "unregistered";
+    private static final String DISABLED = "disabled";
 
     Activity activity;
-    boolean userDefinitelyInvalid = false;
     boolean devServer = true; //If this is a DevAuthBackend server!
+    private LoginActivity context;
+    private boolean userDefinitelyInvalid = false;
 
     public AsyncLogin(Activity activity, String username, String password, boolean devServer) {
         super(ZulipApp.get());
@@ -50,10 +49,9 @@ public class AsyncLogin extends ZulipAsyncPushTask {
 
                 if (obj.getString("result").equals("success")) {
                     this.app.setLoggedInApiKey(obj.getString("api_key"));
-                        if (devServer) ((DevAuthActivity) activity).openHome();
-                        else ((LoginActivity) activity).openHome();
+                    if (devServer) ((DevAuthActivity) activity).openHome();
+                    else ((LoginActivity) activity).openHome();
                     callback.onTaskComplete(result, obj);
-
                     return;
                 }
             } catch (JSONException e) {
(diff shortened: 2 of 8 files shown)


smarx:

app/src/main/java/com/zulip/android/networking/HTTPRequest.java
```diff
@@ -1,141 +1,115 @@
...
         }
     }
 
-    // Java doesn't support a request body on DELETE requests, but RFC2616 does
-    // not prohibit it.
-    class HttpDeleteWithReq extends HttpPost {
-        HttpDeleteWithReq(String url) {
-            super(url);
+    public Response execute(String method, String path) throws IOException {
+        if ("".equals(method)) throw new IOException(app.getString(R.string.method_null));
```
niftynei: does this also check for nulls?

app/src/main/java/com/zulip/android/networking/HTTPRequest.java
```diff
@@ -1,141 +1,115 @@
...
-            httpclient.close();
         }
-
-        if (responseString != null) {
-            Log.i("HTTP.response", responseString);
-        } else {
-            Log.i("HTTP.response", "<empty>");
-        }
-        return responseString;
+        return encodedUrl.substring(0, encodedUrl.length() - 1);
```
niftynei: there's probably a safer way to removing the last character only if it's an `&`.

app/src/main/java/com/zulip/android/networking/ZulipAsyncPushTask.java
```diff
@@ -131,12 +136,12 @@ protected void handleError(Exception e) {
      *
      * @param e the Exception that triggered this handler
      */
-    protected void handleHTTPError(HttpResponseException e) {
-        if (e.getStatusCode() < 500 && e.getStatusCode() >= 400) {
+    protected void handleHTTPError(Exception e) {
+//        if (e.getStatusCode() < 500 && e.getStatusCode() >= 400) {
```
niftynei: If this code is unused, it should be deleted. :)

app/src/main/java/com/zulip/android/networking/ZulipAsyncPushTask.java
```diff
@@ -93,11 +96,14 @@ public void setProperty(String key, String value) {
 
     protected String doInBackground(String... api_path) {
         try {
-            return request.execute(api_path[0], api_path[1]);
-        } catch (HttpResponseException e) {
-            handleHTTPError(e);
+            Response response = request.execute(api_path[0], api_path[1]);
```
niftynei: is it possible that we don't pass enough arguments to this array?

app/src/main/java/com/zulip/android/networking/AsyncLogin.java
```diff
@@ -96,7 +85,7 @@ public void run() {
                 }
             });
             // supermethod invokes cancel for us
-            super.handleHTTPError(e);
+//        super.handleHTTPError(e);
```
niftynei: can remove this if it's unused.

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -107,8 +108,10 @@ public void run() {
                     if (!registeredOrGotEventsThisRun) {
                         request.setProperty("dont_block", "true");
                     }
-                    JSONObject response = new JSONObject(request.execute("GET",
-                            "v1/events"));
+                    Response httpResponse =request.execute("GET",
+                            "v1/events");
+                    String json = httpResponse.body().string();
+                    JSONObject response = new JSONObject(json);
```
niftynei: may need to check the response code for success earlier, since it no longer throws an error.

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -130,9 +133,8 @@ public void run() {
                             }
                         });
                     }
-                } catch (HttpResponseException e) {
-                    if (e.getStatusCode() == 400) {
-                        String msg = e.getMessage();
+                    if (httpResponse.code() == 400) {
```
niftynei: may need to check the response code for success earlier, since it no longer throws an error.

app/src/main/java/com/zulip/android/networking/HTTPRequest.java
```diff
@@ -1,141 +1,115 @@
...
+                    response.body().close();
                     return null;
                 }
             }).execute();
         }
     }
 
-    // Java doesn't support a request body on DELETE requests, but RFC2616 does
-    // not prohibit it.
-    class HttpDeleteWithReq extends HttpPost {
```
niftynei: Here the Delete code extends from POST, but in your updated version, it appears to not support this behavior?

app/src/main/java/com/zulip/android/networking/ZulipAsyncPushTask.java
```diff
@@ -93,11 +96,14 @@ public void setProperty(String key, String value) {
...
+            return response.body().string();
```
niftynei: this might be a good place to check for http error status codes being returned, and calling `handleHTTPError`.

niftynei:
I'm getting this error when running it:

```
06-21 17:21:08.528 E/Error   (18516): oops
06-21 17:21:08.528 E/Error   (18516): org.json.JSONException: Value <html><title>500 of type java.lang.String cannot be converted to JSONObject
06-21 17:21:08.528 E/Error   (18516): 	at org.json.JSON.typeMismatch(JSON.java:111)
06-21 17:21:08.528 E/Error   (18516): 	at org.json.JSONObject.<init>(JSONObject.java:160)
06-21 17:21:08.528 E/Error   (18516): 	at org.json.JSONObject.<init>(JSONObject.java:173)
06-21 17:21:08.528 E/Error   (18516): 	at com.zulip.android.networking.AsyncGetEvents.run(AsyncGetEvents.java:114)
```

niftynei:
Overall this looks good, it needs to handle HTTP errors better (which is what's causing the bug above).

From a UI perspective, it shows me a "Message Sent" toast even though the message isn't actually sent.  This might be better addressed in a separate issue tho.

niftynei:
Note that this is related to issue #29

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -155,6 +161,7 @@ public void run() {
                 } catch (JSONException e) {
                     backoff(e);
                 }
```
kunall17: Without `Thread.sleep` I receive a traceback on my server 
```
AssertionError
2016-06-22 17:00:08,438 INFO     127.0.0.1       GET     429  11ms /api/v1/events (iago@zulip.com via okhttp)
2016-06-22 17:00:08,439 INFO     status=429, data={"msg":"API usage exceeded rate limit, try again in 36.4631140232 secs","result":"error"}
, uid=iago@zulip.com
2016-06-22 17:00:08,441 ERROR    Uncaught exception GET /api/v1/events?queue_id=1466627219%3A4&last_event_id=19& (127.0.0.1)
HTTPRequest(protocol='http', host='localhost:9993', method='GET', uri='/api/v1/events?queue_id=1466627219%3A4&last_event_id=19&', version='HTTP/1.0', remote_ip='127.0.0.1', body='', headers={'Host': 'localhost:9993', 'Client': 'Android', 'Accept-Encoding': 'gzip', 'X-Forwarded-Host': '192.168.100.6:9991', 'Connection': 'close', 'Authorization': 'Basic aWFnb0B6dWxpcC5jb206bVJHRjZ1Nmt4V0RBVHo2NU1VbWJXa3AxUzZrSk9mRG0=', 'User-Agent': 'okhttp/3.2.0'})
Traceback (most recent call last):
  File "/srv/zulip-venv-cache/615b1182e7545b42f90c5ecfcad3168c4ccb6246/zulip-venv/local/lib/python2.7/site-packages/tornado/web.py", line 1042, in _execute
    getattr(self, self.request.method.lower())(*args, **kwargs)
  File "/srv/zulip/zerver/management/commands/runtornado.py", line 184, in get
    self.set_status(response.status_code)
  File "/srv/zulip-venv-cache/615b1182e7545b42f90c5ecfcad3168c4ccb6246/zulip-venv/local/lib/python2.7/site-packages/tornado/web.py", line 245, in set_status
    assert status_code in httplib.responses
 
```

I guess this is because OkHttpClient is running faster than the existing Apache HTTP Clients!
We might have to increase/decrease the sleep time!

kunall17:
@niftynei  Updated the PR, Rebased to master, fixed bugs in sending a private message, added Logs to get the URL and JSON, Added error handling in pushTask!

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -107,8 +109,24 @@ public void run() {
                     if (!registeredOrGotEventsThisRun) {
                         request.setProperty("dont_block", "true");
                     }
-                    JSONObject response = new JSONObject(request.execute("GET",
-                            "v1/events"));
+                    request.setMethodAndUrl("GET", "v1/events");
+                    Response httpResponse = request.execute();
+
+                    String json = httpResponse.body().string();
+                    if (!httpResponse.isSuccessful()) {
```
niftynei: It might be worth also checking the status code?

app/src/main/res/values/strings.xml
```diff
@@ -60,7 +60,7 @@
     <string name="subject_error">No Subject specified.</string>
     <string name="person_error">No Person specified.</string>
     <string name="no_message_error">No Message Written.</string>
-
-
+    <string name="method_null">Method should not be null</string>
+    <string name="method_error">Method Not Supported</string>
```
niftynei: Can you make this lower case? like `Message not supported` ?

app/src/main/java/com/zulip/android/networking/HTTPRequest.java
```diff
@@ -1,141 +1,129 @@
...
+                FormBody.Builder formDeleteBody = new FormBody.Builder();
+                for (Map.Entry<String, String> map : properties.entrySet()) {
+                    formDeleteBody.addEncoded(map.getKey(), map.getValue());
+                }
+                requestBuilder.url(url).delete(formDeleteBody.build());
+                break;
+            case "POST":
+                FormBody.Builder formBody = new FormBody.Builder();
+                for (Map.Entry<String, String> map : properties.entrySet()) {
+                    formBody.addEncoded(map.getKey(), map.getValue());
```
niftynei: I'm running into a problem when logging in with an email.  This API `addEncoded` is transforming the `@` into the encoded version `%40`, making it such that I can't log in. :(

niftynei:
Cool. It's working a lot better now 👍 .

I encountered two problems tho:
1) Logging in with an incorrect username & password doesn't show an error message. 😞   
2) commented in-line as well, but the form encoding is making it such that emails aren't being passed up to the server correctly.  maybe we can turn URI encoding off?

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
kunall17: Before fetching the String by `String json = httpResponse.body().string();` ??

kunall17:
> 1) Logging in with an incorrect username & password doesn't show an error message. :disappointed:

I had to discuss on this one, In this PR if the JSON is recieved and is not successful (status code not lies in 200) then it goes to onTaskFailed in the callback, I think this is more better as the task is failed (task failed for example incorrect login credentials) and previously it went to the onTaskCompleted even if the JSON had a status code on non 200's, which way you think is more liable to be?

This line creates the difference https://github.com/kunall17/zulip-android/blob/e50487c6a94ba8586f9d9181047af2e36e63a803/app/src/main/java/com/zulip/android/networking/ZulipAsyncPushTask.java#L109

app/src/main/java/com/zulip/android/networking/HTTPRequest.java
kunall17: Could not reproduce this problem, I am able to successfully login with the existing code (tried with zulip.tabbott.net as well)
But yeah these strings should not be encoded so I updated the code but it's weird why was it working with my build with the encodedPaths and it works in both the ways, without and with encoding!

niftynei:
@kunall17 Failing the task sounds reasonable to me, as long as you can preserve the error messages appropriately.  I checked this out again though and it still doesn't show an error message with incorrect credentials.  Is this supposed to be fixed?

Managed to log in with out a problem this time, thanks for fixing that encoding error! 👍 

If you can fix up the error messages & rebase this to master (the DevBackends code is still using the old style HTTP stuff), I'll get this merged. Thanks!

kunall17:
@niftynei yeah I left it purposely here, I will complete this and rebase it!

kunall17:
@niftynei updated the code!

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
kunall17: @timabbott After changing from Apache HTTP to okHttpClient asyncGetEvents for the queue works a lot faster and hence it floods the server, so currently I gave a 1 sec pause after every heartbeat receives, will this be the right interval time or increasing this will also do fine? 
The current app seems fine though in receiving new messages!

timabbott: What we do in the web UI is this sort of exponential backoff if the client receives an error from the server (and no backoff other than the fact that the client uses longpolling (aka doesn't pass `dont_block=True` and thus is rate-limited by new events showing up at the server if it receives a success response).  Here's the code for the exponential backoff size:  https://github.com/zulip/zulip/blob/master/static/js/server_events.js#L353

Note that in this case we also show a user-facing error message overlay over the website that explains the client can't connect to the server and with a "try now" button to avoid the user having to wait up to 90s in the event that they've been off the network for a while and just got back on.  

You can see how this looks by logging into a Zulip development server and then running ctrl-C on run-dev.py and watching.

(Since this is a complicated explanation, let me know if you have any questions!)

PR #79, opened 2016-06-28
Fix the previous chatBox espresso test's

kunall17:
015d486: Fixes wrong hint shown in the chatBox

### DIFF
--- a/app/src/main/res/layout/main.xml
+++ b/app/src/main/res/layout/main.xml
@@ -42,7 +42,7 @@
                 android:layout_width="0dp"
                 android:layout_height="wrap_content"
                 android:dropDownWidth="fill_parent"
-                android:hint="@string/stream_txt"
+                android:hint="@string/subject"
                 android:maxLines="1" />
 
             <ImageView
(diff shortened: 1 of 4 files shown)


smarx:

app/src/main/res/layout/main.xml
```diff
@@ -42,7 +42,7 @@
                 android:layout_width="0dp"
                 android:layout_height="wrap_content"
                 android:dropDownWidth="fill_parent"
-                android:hint="@string/stream_txt"
+                android:hint="@string/subject"
```
kunall17: No, it was just a small fixup so I posted this commit here!

niftynei: Ah ok, thanks!

niftynei:
These compile & run but none of them pass. I think the problem is that I'm logged in when they start. Can you write up some docs on what state your app should be in when these start?  

Functional tests are flakey, I'm a bigger proponent of unit tests for Android stuff.

niftynei:
This has been merged!

PR #85, opened 2016-07-12
Implemented Expandable ListView in the streamsDrawer 

kunall17:
This PR changes the current listView used in the streamsDrawer and insert child View's to display the corresponding Subjects/Topics to the related streams 

**Preview:** 
expandlistview

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -48,6 +48,7 @@ import android.widget.AdapterView;
 import android.widget.AdapterView.OnItemClickListener;
 import android.widget.AutoCompleteTextView;
 import android.widget.EditText;
+import android.widget.ExpandableListView;
 import android.widget.FilterQueryProvider;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
@@ -98,6 +99,7 @@ public class ZulipActivity extends AppCompatActivity implements
 
     private DrawerLayout drawerLayout;
     private ActionBarDrawerToggle drawerToggle;
+    ExpandableListView streamsDrawer;
 
     private HashMap<String, Bitmap> gravatars = new HashMap<>();
 
@@ -377,6 +379,48 @@ public class ZulipActivity extends AppCompatActivity implements
         setUpAdapter();
         streamActv.setAdapter(streamActvAdapter);
         topicActv.setAdapter(subjectActvAdapter);
+        checkAndSetupStreamsDrawer();
+    }
+
+    public void setupListViewAdapter() {
+        ExpandableStreamDrawerAdapter streamsDrawerAdapter = null;
+        Callable<Cursor> streamsGenerator = new Callable<Cursor>() {
+            @Override
+            public Cursor call() throws Exception {
+                return ((AndroidDatabaseResults) app.getDao(Stream.class)
+                        .queryBuilder().selectRaw("rowid _id", "name", "color")
+                        .orderByRaw(Stream.NAME_FIELD + " COLLATE NOCASE")
+                        .where().eq(Stream.SUBSCRIBED_FIELD, true).queryRaw()
+                        .closeableIterator().getRawResults()).getRawCursor();
+            }
+        };
+        String[] groupFrom = {Stream.NAME_FIELD, Stream.COLOR_FIELD};
+        int[] groupTo = {R.id.name, R.id.stream_dot};
+        // Comparison of data elements and View
+        String[] childFrom = {Message.SUBJECT_FIELD};
+        int[] childTo = {R.id.name_child};
+        final ExpandableListView streamsDrawer = (ExpandableListView) findViewById(R.id.streams_drawer);
+        streamsDrawer.setGroupIndicator(null);
+        try {
+            streamsDrawerAdapter = new ExpandableStreamDrawerAdapter(this, streamsGenerator.call(),
+                    R.layout.stream_tile_new, groupFrom,
+                    groupTo, R.layout.stream_tile_child, childFrom,
+                    childTo);
+        } catch (Exception e) {
+            ZLog.logException(e);
+        }
+        streamsDrawer.setAdapter(streamsDrawerAdapter);
+    }
+
+    public void checkAndSetupStreamsDrawer() {
+        try {
+            if (streamsDrawer.getAdapter().getCount() != 0) {
+                return;
+            }
+            setupListViewAdapter();
+        } catch (NullPointerException npe) {
+            setupListViewAdapter();
+        }
     }
 
     private void sendMessage() {
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -137,32 +137,6 @@ public class ZulipActivity extends AppCompatActivity implements
         RESET_DATABASE,
     }
 
-    private SimpleCursorAdapter.ViewBinder streamBinder = new SimpleCursorAdapter.ViewBinder() {
-
-        @Override
-        public boolean setViewValue(View arg0, Cursor arg1, int arg2) {
-            switch (arg0.getId()) {
-                case R.id.name:
-                    TextView name = (TextView) arg0;
-                    name.setText(arg1.getString(arg2));
-                    //Change color in the drawer if this stream is inHomeView only.
-                    if (!Stream.getByName(app, arg1.getString(arg2)).getInHomeView())
-                        name.setTextColor(ContextCompat.getColor(ZulipActivity.this, android.R.color.tertiary_text_light));
-                    else name.setTextColor(ContextCompat.getColor(ZulipActivity.this, android.R.color.primary_text_light));
-                    return true;
-                case R.id.stream_dot:
-                    // Set the color of the (currently white) dot
-                    arg0.setVisibility(View.VISIBLE);
-                    arg0.getBackground().setColorFilter(arg1.getInt(arg2),
-                            Mode.MULTIPLY);
-                    return true;
-                default:
-                    break;
-            }
-            return false;
-        }
-
-    };
 
     public HashMap<String, Bitmap> getGravatars() {
         return gravatars;
@@ -209,7 +183,6 @@ public class ZulipActivity extends AppCompatActivity implements
         }
     };
 
-    protected RefreshableCursorAdapter streamsAdapter;
     protected RefreshableCursorAdapter peopleAdapter;
 
     @Override
@@ -233,10 +206,6 @@ public class ZulipActivity extends AppCompatActivity implements
 
     public RefreshableCursorAdapter getPeopleAdapter() {
         return peopleAdapter;
-    }
-
-    public RefreshableCursorAdapter getStreamsAdapter() {
-        return streamsAdapter;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


smarx:

timabbott:
Just wanted to say this looks awesome!

kunall17:
@timabbott This does have a problem, currently I'm extracting the subjects/topics from the received messages but at the initial stage not many messages are received hence, there are less number of topics in the database than the server.

Is there an API to fetch the topics separately or any other possible way to solve this problem?

kunall17:
@timabbott thanks! :) 
And this is the initial implementation to display the unread counts, in a further updates I will update them to display the unread counts in the right space left in the listView.
So how to determine the unread counts of messages, any API routes?

timabbott:
There isn't an API for that, but we could consider adding one.  The other thing we could do is consider fetching more messages in the initial batch when the client starts up.  How many messages does the app currently fetch?  For the webapp, the (compressed) fetch is about 167KB for 1000 messages (which is plenty of history); I wonder if that's small enough that it could be the simplest solution...

timabbott:
For unread counts, there's no convenient API; the current mechanism on web is for the frontend to just compute that data after fetching the message history.  I guess that would be another argument in favor of just trying to fetch data how the website does (if it's fast enough, which is a real concern)

kunall17:
@timabbott Currently the `AsyncGetOldMessages` recieves 101 messages which was of 56.75 KB and mostly the messages contained random one line chat messages, the size is very big than the JSON received in the web UI, api route used is `"v1/messages"` to fetch the old messages.

I have uploaded the JSON I receive.
[messages.txt](https://github.com/zulip/zulip-android/files/360856/messages.txt)

But fetching 1000 messages can be inaccurate as well because there can be more subjects in the remaining messages!

timabbott:
The sizes I was giving are gzip-compressed sizes, not the uncompressed sizes.

It's possible you're doing this already, but if not, you should make sure your API requests are sending the appropriate headers to indicate that you can accept gzip-compressed content (on web the server and client do the compression transparently...).

kunall17:
Well I'm not thorough with the compression of okHttp Network library but it supports _"Transparent GZIP shrinks download sizes."_  as stated in their [website](http://square.github.io/okhttp/)  

I checked the response headers of the POST request of the `"v1/messages"` and I see a header `Transfer-Encoding:chunked`,

Then I added a header `"Accept-Encoding: gzip"` in the POST request no new headers were there in the response headers! 

@niftynei can you help in this thing, please?

niftynei:
@kunall17 

it looks like OkHTTP does Gzip without any prompting: https://github.com/square/okhttp/blob/988142cfe555547bc1df4370babe12c6be86cf3e/okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java#L78-L82

This thread might have some information that's useful: https://github.com/square/okhttp/issues/2132

As a last resort, there's always searching thru the OkHttp source: https://github.com/square/okhttp/search?utf8=%E2%9C%93&q=gzip

Let me know if that doesn't help out.

kunall17:
Yeah I had read this thread, anyways I think to test whether compression is working or not, I should test the size of JSON recieved or the total network usage after disabling the gzip here https://github.com/zulip/zulip/blob/master/puppet/zulip/files/nginx/nginx.conf#L31

Will do this test tomorrow!

app/src/main/java/com/zulip/android/ExpandableStreamDrawerAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+                    "WHERE streams.id=" + groupCursor.getInt(0)).getResults();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        MatrixCursor matrixCursor = new MatrixCursor(new String[]{"subject", "_id"});
+
+        for (String[] result : results) {
+            try {
+                matrixCursor.addRow(new String[]{result[0], groupCursor.getInt(0) + ""});
+            } catch (Exception e) {
```
niftynei: it's a good idea to be explicit about what types of exceptions you're catching.

app/src/main/java/com/zulip/android/ExpandableStreamDrawerAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+        try {
+            results = ZulipApp.get().getDao(Message.class).queryRaw("SELECT DISTINCT subject FROM messages " +
+                    "JOIN streams ON streams.id=messages.stream " +
+                    "WHERE streams.id=" + groupCursor.getInt(0)).getResults();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        MatrixCursor matrixCursor = new MatrixCursor(new String[]{"subject", "_id"});
+
+        for (String[] result : results) {
```
niftynei: If results is null, this will throw an exception.  The best thing to do would be to initialize the `results` to an empty List.

app/src/main/java/com/zulip/android/ExpandableStreamDrawerAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public Cursor getChildrenCursor(Cursor groupCursor) {
+        List<String[]> results = null;
+        try {
+            results = ZulipApp.get().getDao(Message.class).queryRaw("SELECT DISTINCT subject FROM messages " +
+                    "JOIN streams ON streams.id=messages.stream " +
+                    "WHERE streams.id=" + groupCursor.getInt(0)).getResults();
+        } catch (Exception e) {
```
niftynei: it's a good idea to be explicit about what types of exceptions you're catching.
 Add a line note

app/src/main/java/com/zulip/android/ExpandableStreamDrawerAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+            e.printStackTrace();
```
niftynei: We should use `ZLog` or `Log`

app/src/main/java/com/zulip/android/ExpandableStreamDrawerAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+        MatrixCursor matrixCursor = new MatrixCursor(new String[]{"subject", "_id"});
+
+        for (String[] result : results) {
+            try {
+                matrixCursor.addRow(new String[]{result[0], groupCursor.getInt(0) + ""});
+            } catch (Exception e) {
+                e.printStackTrace();
```
niftynei: See note above about printing things.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -446,6 +385,99 @@ public void onClick(View v) {
...
+        String[] childFrom = {Message.SUBJECT_FIELD};
+        int[] childTo = {R.id.name_child};
+        final ExpandableListView streamsDrawer = (ExpandableListView) findViewById(R.id.streams_drawer);
+        streamsDrawer.setGroupIndicator(null);
+        try {
+            streamsDrawerAdapter = new ExpandableStreamDrawerAdapter(this, streamsGenerator.call(),
+                    R.layout.stream_tile_new, groupFrom,
+                    groupTo, R.layout.stream_tile_child, childFrom,
+                    childTo);
+        } catch (Exception e) {
```
niftynei: See above comments about catching exceptions & printing to the logs.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -446,6 +385,99 @@ public void onClick(View v) {
...
+        streamsDrawer.setOnChildClickListener(new ExpandableListView.OnChildClickListener() {
+            @Override
+            public boolean onChildClick(ExpandableListView parent, View v, int groupPosition, int childPosition, long id) {
+                switch (v.getId()) {
+                    case R.id.name_child:
+                        String streamName = ((Cursor) streamsDrawer.getExpandableListAdapter().getGroup(groupPosition)).getString(1);
+                        String subjectName = ((TextView) v).getText().toString();
+                        onNarrow(new NarrowFilterStream(streamName, subjectName));
+                        onNarrowFillSendBoxStream(streamName, subjectName);
+                        break;
```
niftynei: including a default case is best practice, even if it's empty.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -446,6 +385,99 @@ public void onClick(View v) {
...
+        streamsDrawerAdapter.setViewBinder(new SimpleCursorTreeAdapter.ViewBinder() {
+            @Override
+            public boolean setViewValue(View view, Cursor cursor, int columnIndex) {
+                switch (view.getId()) {
+                    case R.id.name:
+                        TextView name = (TextView) view;
+                        final String streamName = cursor.getString(columnIndex);
+                        name.setText(streamName);
+                        //Change color in the drawer if this stream is inHomeView only.
+                        if (!Stream.getByName(app, streamName).getInHomeView())
```
niftynei: if / else statements should always have brackets `{}`

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -446,6 +385,99 @@ public void onClick(View v) {
...
+                return false;
+
+            }
+        });
+        streamsDrawer.setAdapter(streamsDrawerAdapter);
+    }
+
+    public void checkAndSetupStreamsDrawer() {
+        try {
+            if (streamsDrawer.getAdapter().getCount() != 0) return;
```
niftynei: this is potentially confusing, please add brakcet to the `if` statement.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -846,6 +878,18 @@ public void onNarrowFillSendBox(Message message) {
...
     }
+
+    @Override
+    public void onNarrowFillSendBoxStream(String stream, String subject) {
+        switchToStream();
+        streamActv.setText(stream);
+        topicActv.setText(subject);
+        if ("".equals(subject)) {
+            topicActv.requestFocus();
+        } else messageEt.requestFocus();
```
niftynei: please use `{}` brackets on if - else statements.

app/src/main/java/com/zulip/android/ExpandableStreamDrawerAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+                    "JOIN streams ON streams.id=messages.stream " +
+                    "WHERE streams.id=" + groupCursor.getInt(0)).getResults();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        MatrixCursor matrixCursor = new MatrixCursor(new String[]{"subject", "_id"});
+
+        for (String[] result : results) {
+            try {
+                matrixCursor.addRow(new String[]{result[0], groupCursor.getInt(0) + ""});
```
niftynei: `String.valueOf(groupCursor.getInt(0))` is better than using `+ ""`.

niftynei:
hey @kunall17  this looks really good!

There's 1 bug that's a blocker for merge: clicking on a stream header no longer narrows to just that stream.  This makes it impossible to browse through all of the topics in a stream.  It also means that if a stream has no topics listed, it's impossible to narrow to that stream.

Another small thing I noticed:  muted streams subtopics are not greyed out in color (they're black).  It'd be nice if they were the same font color as the stream is (a lighter grey).

kunall17:
@niftynei updated..:)

niftynei:
@kunall17 thanks kunal! Did you have a chance to implement on the top level stream button yet?

kunall17:
@niftynei top level stream button, can you elaborate little more please??

niftynei:
Sure! For example, on our zulip instance this stream doesn't have any topics under it. When I tap on it, it doesn't seem to do anything (in the current client it narrows to the AMA stream).

screenshot_20160719-190332

niftynei:
cc @kunall17

kunall17:
Yeah, currently it does nothing, so you want to if there are no subtopics/subjects it should narrow to the stream?

niftynei:
Yeah, it should still always narrow to stream (even if there are sub-topics to expand).

kunall17:
decba2e: narrows on group click, btw we could click on the textView to narrow it separately as well!

niftynei:
Cool. Looks good to me as is. 👍

niftynei:
I realize this doesn't include un-read counts -- a separate PR for that is good :)

kunall17:
yeah, that will be in another PR!

PR #99, opened 2016-07-20
Removed Use zulip.com checkbox

kunall17:
Fixes half of #88 
optimize flow for entering server domain in another PR!

PR #101, opened 2016-07-26
Better Login Flow

kunall17:
Uses `v1/get_auth_backends` to get backends and show the controls for available AuthBackends for the specified server!
Also uses commit 2e15b26 (Removed Use zulip.com checkbox) from #99 

Preview: 
ezgif-2264385963


niftynei:
@kunall17 I entered in the zulip.com domain, but the "Login" button did not appear. I can put in my user name & password, but there's no button to push to login.

Also the `Legal information` button doesn't appear to do anything.

screenshot_20160728-155218

kunall17:
@niftynei fixed the little bug!

niftynei:
Cool, thanks @kunall17 !

PR #107, opened 2016-07-27
Add links to Google Play store.

brainwane:
Adding links for beta channel and for current released app.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,12 @@
 Zulip for Android
 =================
 
+This is the Android client for `Zulip <http://www.zulip.org>`_.
+It is available `in the Google Play store 
+<https://play.google.com/store/apps/details?id=com.zulip.android>`_.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -2,6 +2,12 @@
 Zulip for Android
 =================
 
+This is the Android client for `Zulip <http://www.zulip.org>`_.
+It is available `in the Google Play store 
+<https://play.google.com/store/apps/details?id=com.zulip.android>`_.
+You can also `join our beta channel <https://play.google.com/apps/testing/com.zulip.android>`_ 
+after which the Play store should auto-update you.
```
timabbott: Would probably be a bit more verbose about how the beta may be buggy, etc.

timabbott: Also, maybe we should post a comment on #1 about the beta channel app having that feature

Issue #111, opened 2016-07-29
Set user-agent properly for requests to the server from the Android app


timabbott:
Based on server logs, it appears this regressed with the switch to the okhttp library, I think.

timabbott:
(In theory we probably want the user-agent to be something nice indicating it's the Zulip Android app and what version of the Zulip Android app it is)

timabbott:
Fixed in #113.  @kunall17 FYI if you put `Fixes: #113` in one of the commits in a pull request, the issue will be automatically closed when the PR is merged.

PR #113, opened 2016-08-02
Add User-Agent header to HTTPRequest

kunall17:
Implement #111 
Adds header like ZulipAndroid-0.8.9 
And upgrade okHttp library dependency

### DIFF
--- a/app/src/main/java/com/zulip/android/ZulipApp.java
+++ b/app/src/main/java/com/zulip/android/ZulipApp.java
@@ -35,7 +35,7 @@ public class ZulipApp extends Application {
     public static final String API_KEY = "api_key";
     public static final String EMAIL = "email";
     private static ZulipApp instance;
-    private static final String USER_AGENT = "ZulipMobile";
+    private static final String USER_AGENT = "ZulipAndroid";
     private static final String DEFAULT_SERVER_URL = "https://api.zulip.com/";
     private Person you;
     private SharedPreferences settings;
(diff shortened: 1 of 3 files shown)


smarx:

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -158,12 +158,12 @@ public String getApiKey() {
     public String getUserAgent() {
         try {
             return ZulipApp.USER_AGENT
-                    + "/"
+                    + "-"
```
niftynei: This should stay as a `/`.  User agents are usually in the form `software/version`

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -158,12 +158,12 @@ public String getApiKey() {
...
                     + getPackageManager().getPackageInfo(getPackageName(), 0).versionName;
         } catch (NameNotFoundException e) {
             // This should… never happen, but okay.
             ZLog.logException(e);
-            return ZulipApp.USER_AGENT + "/unknown";
+            return ZulipApp.USER_AGENT + "-unknown";
```
niftynei: Same as above comment. This should stay as a slash.

niftynei:
See comments about `/` vs `-`.  

Otherwise lgtm

niftynei:
Thanks @kunall17 !

Issue #112, opened 2016-08-01
Notification support


pkaske:
It seems there is no way to have notifications when new messages arrive. Am I overlooking something?

timabbott:
The Android app supports push notifications, but I think there may be some tooling complexities with doing this against your own server.  I've been meaning to try to find time to look into how this works.  @kunall17 @niftynei is your understanding that GCM is tied to the Google Auth tangle of issues around having a single app that can talk to multiple servers run by different parties?

kunall17:
Yeah the current google ID is also hardcoded https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/gcm/Notifications.java#L25 in the application, well I tried to debug if the dev server is sending GCM messages I never get these logging error or info here  [L185](https://github.com/zulip/zulip/blob/0aae0eab497f7a770fa8f17b32fc471405f6228e/zerver/lib/push_notifications.py#L185) or [L195](https://github.com/zulip/zulip/blob/0aae0eab497f7a770fa8f17b32fc471405f6228e/zerver/lib/push_notifications.py#L195) or any other GCM debug! 

But why do we need GCM, we get the new messages via the AsycGetEvents queue?

kunall17:
EDIT: GCM should be used, now got it! 

I'll post a PR to fetch the google client ID and register from that ID.

niftynei:
afaict GCM has been working great, both on dev and prod builds.

kunall17:
@timabbott have you checked if the GCM is sending messages from the server side?

timabbott:
Well for zulip.tabbott.net we currently get this server error logged, so maybe that server needs some configuration done.
`Attempting to send a GCM push notification, but no API key was configured`


@kunall17 what does one need to do to enable GCM on the server side with our current setup?  (We should add how to do that to our documentation, possibly for both the server and the app, but I guess we should start with my setting it up for zulip.tabbott.net)

pkaske:
Doc on that would be awesome!

kunall17:
@timabbott I changed the ID in the android project and tried to register a GCM in the server, can you check?

I'll add docs once we get this working!

timabbott:
There's exactly one such GCM warning in the logs from today; :
`2016-08-04 15:51:32,528 ERROR    Attempting to send a GCM push notification, but no API key was configured`
[...]

PR #114, opened 2016-08-05
Save Client ID for registering GCM to that client

kunall17:
As in https://github.com/zulip/zulip-android/issues/112#issuecomment-236796320 This now fetches Google client ID and uses that one to register a GCM client ID

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -84,6 +84,7 @@ import com.zulip.android.models.Presence;
 import com.zulip.android.models.PresenceType;
 import com.zulip.android.R;
 import com.zulip.android.models.Stream;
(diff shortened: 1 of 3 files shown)


smarx:

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -237,8 +239,7 @@ protected void onCreate(Bundle savedInstanceState) {
             return;
         }
         this.logged_in = true;
-        notifications = new Notifications(this);
-        notifications.register();
+        registerGCM();
```
niftynei: Needs to 'unregister' GCM after logout, including delete the saved google_client_id preference .

niftynei:
looks pretty good, see comments in line.  In my test I didn't receive any notifications when I was pinged.

niftynei:
Superceded by #123

Issue #136, opened 2016-08-17
Narrowing by touching a stream in message view doesn't work


timabbott:
In the main message view, if you touch a topic, it nicely narrows to the stream+topic pair for the message.  However, if you touch a stream, it (incorrectly) reports that there are no messages in that narrow.

timabbott:
Also, and perhaps relatedly, touching a PM sender to narrow to messages with that person crashes the app.

PR #139, opened 2016-08-18
Fixes: #136 Narrow by stream on click

kunall17:
https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/filters/NarrowFilterStream.java#L58 here checks for null and not "" therefore this problem was arising!


smarx:

timabbott:
This looks good.  I think a follow-up thing we'll want to do is ensure the narrowing is done in a way where the message that you touched to initiate the narrow on stays in view (rather than jolting around).  See for notes on how this works in the webapp.

http://zulip.readthedocs.io/en/latest/pointer.html

niftynei:
lgtm. closing this PR!

Issue #132, opened 2016-08-17
Replying to messages that you sent to yourself doesn't work


timabbott:
(1) Send yourself a message
(2) Press it in the Android app to try to reply -- it doesn't fill in the recipient bar.

timabbott:
This seems to be still broken as well.

niftynei:
in progress.

PR #140, opened 2016-08-18
Fixes: #132 Show Title and name in header if narrowed to self PM's

kunall17:
I have also checked this function https://github.com/kunall17/zulip-android/blob/43fef77636256b2f7ec51d560f0c7f6aa35be88c/app/src/main/java/com/zulip/android/models/Message.java#L241 which is being edited in this PR, is only used in this binding the Recipent Bar.


smarx:

timabbott:
The bug I reported only impacted messages you sent directly to yourself; this patch seems to change the rendering of group conversations between you and other people in a way that I found confusing:

image

kunall17:
Fixed logged in as Iago in the screenshot!
device-2016-08-20-154726

niftynei:
lgtm. fixes issue shown in @timabbott 's screenshots.

PR #141, opened 2016-08-21
Espresso Test 

kunall17:
This creates espresso tests for the UI.

PR #142, opened 2016-08-21
Fixes #131: Wrong messages order 

### LINKED ISSUE #131: Messages are out of order
I've noticed that since the new look and feel update, messages are often times getting inserted in the wrong order. For example when narrowed to a private chat, 10 or so older messages are appearing at the bottom of the list, below newer messages.  The timestamps on these messages are correct, but how they're ordered is not.

### LINKED ISSUE #132: Replying to messages that you sent to yourself doesn't work

kunall17:
Currently [this](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L163) was adding messages before another MessageHeaderParent now this thing is fixed and comparing ID's now! 

Commit 2ef8d48: Shows all the messages of today at the bottom as new messages even if the app pointer is greater than this messageId.

The [code used ](https://github.com/zulip/zulip-android/pull/141/files#diff-1dded4e9deaa29c6eb8becf0bb2fee03R54) in creating [espresso test](https://github.com/zulip/zulip-android/pull/141) were very useful I used them to debug the order + if they are included in the correct messageHeaders. 

I have created a [**temporary patch**](https://github.com/zulip/zulip-android/files/428902/checkOrder.and.messageHeaderParent.zip) for espresso code just for debugging purpose of this branch what this does is it invokes both the methods to check the message order according to ID's and second one checks if the messages are inserted below the correct messageHeader.
This works if we go to overflow menu and press the Refresh menuItem.
If throws some errors in the log then it means there are still bugs! else it will print the information about the Id's and headerParents ID's

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
+++ b/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
@@ -343,7 +343,7 @@ public class MessageListFragment extends Fragment implements MessageListener {
         int topPosBefore = linearLayoutManager.findFirstVisibleItemPosition();
         int addedCount = 0;
         int headerParents = 0;
-
+        StringBuilder stringBuilder = new StringBuilder ();
         if (pos == LoadPosition.NEW && !loadedToBottom) {
             // If we don't have intermediate messages loaded, don't add new
             // messages -- they'll be loaded when we scroll down.
@@ -383,7 +383,7 @@ public class MessageListFragment extends Fragment implements MessageListener {
                 this.adapter.addNewMessage(message);
                 messageList.add(message);
             } else if (pos == LoadPosition.ABOVE || pos == LoadPosition.INITIAL) {
-                headerParents = (this.adapter.addMessage(message, addedCount + headerParents)) ? headerParents + 1 : headerParents;
+                headerParents = (this.adapter.addOldMessage(message, addedCount + headerParents, stringBuilder)) ? headerParents + 1 : headerParents;
                 messageList.add(addedCount, message);
                 addedCount++;
             }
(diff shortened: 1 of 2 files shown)


smarx:

niftynei:
Hey @kunall17, this seems to work for messages sent today, but any past dates are out of order across message thread boundaries. 


In this screenshot, dates are out of order.
screenshot_20160822-143554

In this screenshot, times are out of order.
screenshot_20160822-143655

koziodigitalinc:
@niftynei Yeah this is what my concern was while reviewing this pull request. I pulled down the branch to examine the the code.

kunall17:
Will look into this!

kunall17:
This thing even confused me, well the branch is working fine and does what it is coded to be! 

In the second screenshot you can see the first message which is at "Aug 20, 06:25" has been added.
After that a new message (timestamp "Aug 20, 06:30") with different messageHeader has been added now after that the two messages at "Aug 20, 08:34" have been added to same old messageHeader this way the old messages are displayed, which is causing the confusion!

niftynei:
hi @kunall17. thanks for looking into this. the coded functionality, even if working as intended, is incorrect. The order which messages are displayed should exactly mirror that of the web app -- where all messages are globally ordered by their timestamp, or ID.  Right now the app re-arranges messages by 'topic' -- putting them out of order 'globally'.

kunall17:
@niftynei Fixed this branch, now this has the same message order as the web UI.
Sorry for the confusion!

niftynei:
Thanks @kunall17 !

PR #143, opened 2016-08-22
Implemented swipe to remove chatBox

kunall17:
record


smarx:

niftynei:
nice!  works well afaict.  it's not the most discoverable functionality, but it definitely serves it's purpose. 👌

kunall17:
Agreed, but this will be needed as after narrowing it will focus the message edit text and then the chat box won't go automatically on scrolling, so then hopefully user will try to swipe to remove chatBox

niftynei:
💯  🌋

Issue #187, opened 2016-09-30
Default HTTP Scheme used in non prod builds is http


kunall17:
Sometimes (like connecting to zulip.tabbott.net) the app currently uses default scheme as http if not specified which is wrong, it ask the developer to which scheme to use!

Issue #185, opened 2016-09-25
Cannot login through EmailBackend 


kunall17:
After the RetroFit library in the app, the login through E-Mail is not successful.
I tried with the zulip.tabbott.net 

The full Logcat
```
09-25 12:56:16.026 21284-21284/com.zulip.android.dev W/System: ClassLoader referenced unknown path: /data/user/0/com.google.android.gms/app_chimera/m/00000007/n/arm64-v8a
09-25 12:56:16.163 21284-21284/com.zulip.android.dev D/AppTracker: App Event: start
09-25 12:56:16.183 21284-21388/com.zulip.android.dev D/OpenGLRenderer: Use EGL_SWAP_BEHAVIOR_PRESERVED: true
09-25 12:56:16.224 21284-21284/com.zulip.android.dev D/PhoneWindow: state.preset = true, color: -16777216, translucent = 0
09-25 12:56:16.284 21284-21388/com.zulip.android.dev I/Adreno: QUALCOMM build                   : ccaad82, Ic3eab96f3e
                                                               Build Date                       : 07/18/16
                                                               OpenGL ES Shader Compiler Version: XE031.07.00.01
                                                               Local Branch                     : 
                                                               Remote Branch                    : refs/tags/AU_LINUX_ANDROID_LA.HB.1.3.1.C1.06.00.01.237.007
                                                               Remote Branch                    : NONE
                                                               Reconstruct Branch               : NOTHING
09-25 12:56:16.290 21284-21388/com.zulip.android.dev I/OpenGLRenderer: Initialized EGL, version 1.4
09-25 12:56:16.324 21284-21284/com.zulip.android.dev D/AbstractTracker: Event success
09-25 12:56:16.345 21284-21388/com.zulip.android.dev E/HAL: load: gralloc.msm8996.so
09-25 12:56:16.615 21284-21284/com.zulip.android.dev E/SpannableStringBuilder: SPAN_EXCLUSIVE_EXCLUSIVE spans cannot have a zero length
09-25 12:56:16.616 21284-21284/com.zulip.android.dev E/SpannableStringBuilder: SPAN_EXCLUSIVE_EXCLUSIVE spans cannot have a zero length
09-25 12:56:16.965 21284-21284/com.zulip.android.dev D/AppTracker: App Event: stop
09-25 12:56:17.046 21284-21284/com.zulip.android.dev D/AbstractTracker: Event success
[...]

PR #188, opened 2016-09-30
Fixes: #187 Ask for HTTP/HTTPS for non-prod builds

kunall17:
Fix for #185 
A small fix for the HTTP/HTTPS default in debug mode! 
This asks the user if HTTP/HTTPS scheme if not specified in non-prod builds!


smarx:

timabbott:
I would prefer to do something that does auto-detection and thus doesn't involve an extra prompt, but given that this issue is making life difficult for folks trying to contribute, I've gone ahead and merged this.  We can always improve it from here.

Issue #181, opened 2016-09-14
Update read flags and pointer when reading messages


timabbott:
Logically:
- any message the user sees the entirety of should be being marked as read via the `POST messages/flags` endpoint.  
- If the user scrolls down in the home view, the latest message they saw in the home view, if newer than the user's pointer, should be submitted to the server as an update to the pointer via `PUT users/me/pointer`.

Sam1301:
Hi, I'd like to work on this issue.

brainwane:
Cool! @Sam1301 Go ahead. If you need any help, please speak up -- https://blogs.akamai.com/2013/10/you-must-try-and-then-you-must-ask.html might be a good guide. :)

kunall17:
@timabbott 
File https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/networking/AsyncUnreadMessagesUpdate.java  uses the `messages/flags` endpoint and 

https://github.com/zulip/zulip-android/search?utf8=%E2%9C%93&q=updatePointer&type=Code uses the end point `users/me/pointer`

They are being updated I think, is there any problem?

timabbott:
I've heard reports of there being a problem with this, but I think maybe the problem is that users who are not reading anything in the home view (and instead just look at narrows) don't end up updating the pointer on the server properly.  There's a server-side solution we need to do for that, so we can close this.

Sam1301:
@timabbott @kunall17 I believe we are updating the pointer and read flags. Actually whats left is sync up with server for "read" messages (update_message_flags event) and updating the database accordingly which I have incorporated in my unread counts pr work. 
Should this be closed?

timabbott:
Yeah, I think so.

PR #205, opened 2016-10-11
Update pointer and read flags

Sam1301:
When `filter != null` then , `startedFromFilter` should be true. This will successfully update the pointers in HomeView.
Also, initially `message.getMessageRead` returns `null` and hence throws an exception
> attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()  on a null object reference

Due to this runtime exception, the control never reaches `zulipApp.markMessageAsRead(message)` 
Hence to avoid this, I have used `isMessageRead` boolean which is set to false if `message.getMessageRead()` is null or false and true if `message.getMessageRead()` is true.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
+++ b/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
@@ -380,7 +380,7 @@ public class RecyclerMessageAdapter extends RecyclerView.Adapter<RecyclerView.Vi
     @Override
     public void onViewAttachedToWindow(RecyclerView.ViewHolder holder) {
         super.onViewRecycled(holder);
-        if (holder.getItemViewType() == VIEWTYPE_MESSAGE)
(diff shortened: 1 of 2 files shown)


smarx:

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
```diff
@@ -394,7 +394,12 @@ private void markThisMessageAsRead(Message message) {
             if (!startedFromFilter && zulipApp.getPointer() < mID) {
```
kunall17: Also this `!startedFromFilter` can be moved to L383!

niftynei:
lgtm. is there any open issues that this fixes @Sam1301 ?

niftynei:
merged!

Sam1301:
This issue #181 was closed, but as Kunal and I discussed here https://zulip.tabbott.net/#narrow/near/35157/stream/android/topic/Issue.20.23181 ,it seemed like a genuine bug.

PR #207, opened 2016-10-11
Logout on Unauthorized & Show Login Error Messages

niftynei:
This pull request fixes a few edge case failures such as:
1) Unauthorized users will return to the login screen and
2) Errors with logging in will display the reason for login failure (ie incorrect username & password)


smarx:

Issue #173, opened 2016-09-12
Avatars sometimes shown on green dots inconsistently


timabbott:
It seems some users' avatars aren't being loaded properly in the Android app, and are displayed as a green presence dot, rather than the person's actual avatar, in the actual message feed.

E.g. for me, on zulip.tabbott.net, Sumana's avatar and my avatar don't show up, nor does Rishi's or GitHub Bot's, but everyone else has their avatar shown properly.  

If I had to guess, we had a network error when loading these avatars at some point, and incorrectly cached that "couldn't fetch the avatar" state permanently.  

(As a sidenote, I'm not sure why a green dot would be a good choice as the "couldn't fetch avatar" icon -- would be better to use something with the same form factor as our actual square avatars)

brainwane:
@timabbott Would it be ok to break out the "better 'couldn't fetch avatar' icon" feature as its own issue?

timabbott:
yes, that'd be totally reasonable.

Sam1301:
Working on it :+1:

niftynei:
The bug with avatars not showing up is a data model problem -- for some messages the sender details are not properly filled in. I'm investigating.

PR #211, opened 2016-10-12
Fixup blank avatars and senders

niftynei:
There's been a persistent issue with avatars and sometime sender names appearing as blank in the app.  This fixes both these issues by 

1) Properly serializing the Sender data in the Message object before writing it to the DAO &
2) Prefixing path-only avatar urls with the server domain


smarx:

niftynei:
thanks to @Sam1301 for identifying the root causes of these problems.

niftynei:
this fixes #173 .

PR #212, opened 2016-10-12
Fix a variety of user interface reply-to-sender bugs

niftynei:


smarx:

PR #213, opened 2016-10-12
Fixes #172

### LINKED ISSUE #172: Incoming messages do not reliably show up at bottom of home view feed
I have a user report of multiple times new messages being sent (even sometimes sent by the current user) and not being correctly added at the bottom of the message feed (the message just never appears, even though other ones before and after do).  

This probably a priority bug to track down.

niftynei:
New messages being pushed to the phone weren't appearing at the bottom of the feed; this fixes that issue.


smarx:

PR #217, opened 2016-10-17
Fixes #153: Show in-app new messages notifications snackbar

### LINKED ISSUE #153: Implement a way to notify when a new message is recieved
This is very important for the app, as currently no notifications is given to the user if a new message is recieved when the user scrolls to bottom then only he knows if a new message is recieved.

Suggestion: When in homeList (not narrowed to a topic or stream currently) then we can change the title to Zulip (2) which shows 2 new messages received, and the stream drawer will show which stream/topic received the message by #120.

Complexity: One problem will be while implementing this would be, when we send a message through chatBox by [AsyncSend ](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/networking/AsyncSend.java) java file, after sending this message, the server resend's this message to the android App to save in the database. 

But yeah we can also easily distinguish by checking who is the sender of this message and based on that we can increase the count.

kunall17:
Shows a snackbar in the top, snackbar message text changes according to private/stream messages and also changes text if messages from multiples topics/private messages recieved!

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -2070,13 +2070,20 @@ public class ZulipActivity extends BaseActivity implements
         TypedValue tv = new TypedValue();
         if (getTheme().resolveAttribute(android.R.attr.actionBarSize, tv, true))
             mToolbarHeightInPx = TypedValue.complexToDimensionPixelSize(tv.data, getResources().getDisplayMetrics());
+        snackbar.setCallback(new Snackbar.Callback() {
+            @Override
+            public void onDismissed(Snackbar snackbar, int event) {
+                super.onDismissed(snackbar, event);
+                if (event == Snackbar.Callback.DISMISS_EVENT_TIMEOUT) prevMessageSameCount = -1;
+            }
+        });
     }
 
     NarrowFilter narrowFilter;
-
+    String prevId = null;
+    int prevMessageSameCount = -1;
     private void showSnackbarNotification(Message[] messages) {
         MutedTopics mutedTopics = MutedTopics.get();
-        String prevId = null;
         int nonMutedMessagesCount = 0;
         Message tempMessage = null; //Stores a temporary message which is non-muted, later used for retrieving Stream/Topic
         for (Message message : messages) { //Check if all messages from same topic/private and remove all the muted messages
@@ -2086,13 +2093,16 @@ public class ZulipActivity extends BaseActivity implements
             if (prevId != null && !prevId.equals(message.getIdForHolder())) {
                 prevId = null;
                 tempMessage = null;
+                prevMessageSameCount = 0;
                 break;
+            } else {
+                prevMessageSameCount++;
             }
             prevId = message.getIdForHolder();
             if (tempMessage == null) tempMessage = message;
         }
         if (nonMutedMessagesCount == 0) return;
-        if (prevId == null) {
+        if (prevId == null && messages.length > 1) {
             snackbar.setText(getResources().getQuantityString(R.plurals.new_message_mul_sender, nonMutedMessagesCount, nonMutedMessagesCount));
             narrowFilter = null;
             if (narrowedList != null) {
@@ -2107,7 +2117,9 @@ public class ZulipActivity extends BaseActivity implements
                 }
             });
         } else {
+            if (messages.length == 1) tempMessage = messages[0];
             String name = (tempMessage.getType() == MessageType.PRIVATE_MESSAGE) ? getString(R.string.notify_private, tempMessage.getSenderFullName()) : getString(R.string.notify_stream, tempMessage.getStream().getName() , tempMessage.getSubject());
+            if (prevMessageSameCount > 0) name += " (" + prevMessageSameCount + ")";
             snackbar.setText(getResources().getQuantityString(R.plurals.new_message, nonMutedMessagesCount, nonMutedMessagesCount, name));
             narrowFilter = (tempMessage.getType() == MessageType.PRIVATE_MESSAGE) ? new NarrowFilterPM(Arrays.asList(tempMessage.getRecipients(app))) : new NarrowFilterStream(tempMessage.getStream(), tempMessage.getSubject());
             snackbar.setAction(R.string.SHOW, new View.OnClickListener() {
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -1,6 +1,15 @@
 package com.zulip.android.activities;
 
 import android.Manifest;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.Callable;
+import java.util.ArrayList;
+import java.util.concurrent.TimeUnit;
+
 import android.animation.Animator;
 import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
@@ -29,7 +38,9 @@ import android.os.Environment;
 import android.os.Handler;
 import android.provider.MediaStore;
 import android.support.design.widget.AppBarLayout;
+import android.support.design.widget.CoordinatorLayout;
 import android.support.design.widget.FloatingActionButton;
+import android.support.design.widget.Snackbar;
 import android.support.v4.app.ActionBarDrawerToggle;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.app.FragmentManager;
@@ -47,6 +58,7 @@ import android.text.Editable;
 import android.text.TextUtils;
 import android.text.TextWatcher;
 import android.util.Log;
+import android.view.Gravity;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
@@ -539,6 +551,7 @@ public class ZulipActivity extends BaseActivity implements
         }
         handleOnFragmentChange();
         calendar = Calendar.getInstance();
+        setupSnackBar();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


smarx:

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1545,5 +1550,55 @@ public void onNewMessages(Message[] messages) {
...
+                        FragmentManager.POP_BACK_STACK_INCLUSIVE);
+            }
+            snackbar.setAction(R.string.SHOW, new View.OnClickListener() {
+                @Override
+                public void onClick(View view) {
+                    homeList.showLatestMessages();
+                }
+            });
+        } else {
+            String name = ((messages[0].getType() == MessageType.PRIVATE_MESSAGE) ? "From " + messages[0].getSenderFullName() : " in " + messages[0].getStream().getName() + "\\" + messages[0].getSubject());
```
niftynei: Yep.  You can use string formatting instead of concatenating them, something like:

```
getString("From %s in %s \\ %s", senderFullName(), getName(), getSubject());
```

niftynei:
If possible, can you make the snackbars appear below the ActionBar when it's visible?  In the screenshots it looks like it appears on top of it.

kunall17:
Updated

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1554,14 +1557,16 @@ public void onNewMessages(Message[] messages) {
...
         snackbar = Snackbar.make(coordinatorLayout, "", Snackbar.LENGTH_LONG);
         View v = snackbar.getView();
-        CoordinatorLayout.LayoutParams params = (CoordinatorLayout.LayoutParams) v.getLayoutParams();
-        params.gravity = Gravity.TOP;
-        v.setLayoutParams(params);
+        snackBarParams = (CoordinatorLayout.LayoutParams) v.getLayoutParams();
+        snackBarParams.gravity = Gravity.TOP;
+        v.setLayoutParams(snackBarParams);
+
+        TOOLBAR_HEIGHT_IN_PX = convertDpToPixel(56);
```
niftynei: There's a system resource value that encodes the toolbar height, that should update itself depending on whether the phone is vertical or horizontal.  See http://stackoverflow.com/a/13216807/903680

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1599,6 +1604,19 @@ public void onClick(View view) {
...
         }
+        if (appBarLayout.getVisibility() == View.GONE) {
+            snackBarParams.setMargins(0, 0, 0, 0);
+        } else {
+            snackBarParams.setMargins(0, TOOLBAR_HEIGHT_IN_PX, 0, 0);
+        }
         snackbar.show();
     }
+
+    private static int TOOLBAR_HEIGHT_IN_PX;
```
niftynei: Per Android/Java code style guidelines, only constants (variables that are static & final) are written in all caps. I'd suggest either making this a member variable (removing `static`) or adding `final` to make it a constant.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1599,6 +1604,19 @@ public void onClick(View view) {
...
+
+    private int convertDpToPixel(float dp) {
```
niftynei: This is unnecessary if you use the system property, see comment above.

niftynei:
@kunall17 this looks really great.  I noticed a few issues that aren't commentable via code:
1) The animation for the snackbar is oriented toward the bottom. This makes sense if the snackbar is anchored to a bottom element, but seems really strange since it's anchored to the top.  [Reversing the interpolator](http://stackoverflow.com/a/5259995/903680) may be an easy way to do it. 
2) The notification is showing for topics that I have muted. Since they're muted, I shouldn't hear from them which means I shouldn't see a notification either :)
3) Finally, this is a bit of a stretch goal. When I'm scrolled to the bottom of the feed, I shouldn't see notifications for messages since they will just appear on the screen.

app/src/main/res/values/strings.xml
```diff
@@ -89,4 +89,9 @@
     <string name="no_topic_in_message">(no topic)</string>
     <string name="message_copied">Message Copied</string>
     <string name="login_failed">Unable to login</string>
+    <string name="new_message_mul_sender" formatted="false">%d New Message%s</string>
+    <string name="new_message" formatted="false">%d New Message%s %s</string>
```
niftynei: It's best practice to number the strings that are being replaced, eg `<string name="welcome_messages">Hello, %1$s! You have %2$d new messages.</string>`

See Formatting Strings in https://developer.android.com/guide/topics/resources/string-resource.html

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1545,5 +1550,55 @@ public void onNewMessages(Message[] messages) {
...
+                        FragmentManager.POP_BACK_STACK_INCLUSIVE);
+            }
+            snackbar.setAction(R.string.SHOW, new View.OnClickListener() {
+                @Override
+                public void onClick(View view) {
+                    homeList.showLatestMessages();
+                }
+            });
+        } else {
+            String name = (messages[0].getType() == MessageType.PRIVATE_MESSAGE) ? String.format(Locale.getDefault(), getString(R.string.notify_private), messages[0].getSenderFullName()) : String.format(Locale.getDefault(), getString(R.string.notify_stream), messages[0].getStream().getName() , messages[0].getSubject());
```
niftynei: the built-in function `getString()` will handle string formatting for you.  this will cut down on the length of this line a lot.

in other words you can re-write this:
```
String.format(Locale.getDefault(), getString(R.string.notify_private), messages[0].getSenderFullName())
```

as this
```
getString(R.string.notify_private, messages[0].getSenderFullName())
```

Docs: https://developer.android.com/reference/android/content/Context.html#getString(int, java.lang.Object...)

kunall17:
@niftynei 
> 1) The animation for the snackbar is oriented toward the bottom. This makes sense if the snackbar is anchored to a bottom element, but seems really strange since it's anchored to the top. Reversing the interpolator may be an easy way to fix it. 

Didn't understand this one completely, there is no noticeable default animation given for the snackbar? 

Implemented the rest of them :)

kunall17:
@niftynei I updated this before only! Forgot to post a comment on the update, if you haven't seen this!

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -144,7 +149,7 @@
     private AsyncGetEvents event_poll;
 
     private Handler statusUpdateHandler;
-
+    private int toolbarHeightInPx;
```
Sam1301: This is nitpicking but I think you forgot to rename this variable starting with 'm' (for member variables).

Sam1301:
Can we set gravity for the snackbar at the bottom? In case we can't, creating a custom view with animation from top might be a potential solution. I'd like to know your views on this :)

kunall17:
@Sam1301 well the snackbar is set to display on the top on purpose, just UI/UX preference and to make it like the web UI, as there also the notifications are displayed on top!

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
kunall17: No worries, updated :)

Sam1301:
Thanks, i absolutely love this feature. @kunall17 Can it be merged and animation problem dealt with later?

kunall17:
@Sam1301 check it out now, updated the PR to animate snackbar from top!

kunall17:
This TSnackbar turns out to be very buggy, we should omit the animation changes for now and let it anchor from bottom!

Sam1301:
@kunall17 If you approve, I would like to work on this by creating some sort of custom banner achored at the top.

kunall17:
@Sam1301 yeah, sure I think we should merge this one and if you want you update with some custom view go ahead!
@timabbott LGTM reviewed by niftynei and Sam1301! :)

timabbott:
@kunall17 can you resolve the new merge conflicts?

timabbott:
Also can you post a screenshot of the current version just so I know what it looks like?

kunall17:
Sure, update the branch! :)

ezgif com-f05b3b87f3

saketkumar95:
@kunall17 How about swipe to right or left to remove the notification?

kunall17:
@saketkumar95 Yeah it would be good for a future update, its good for now!
@timabbott updated the branch! :)

timabbott:
@saketkumar95 @Sam1301 can one of you give the new version a careful review?

Sam1301:
Sure :+1:

saketkumar95:
@timabbott yeah sure. :+1:

Sam1301:
@timabbott The use of erroneous `TSnackBar` has been changed, so I guess the recent changes were just reverted back. 

- On clicking _show_ in the in-app notification bar, the stream is narrowed.
- Notifications are caught in homeView as well as in streams for all messages.

Here's a video for how the notification looks
https://drive.google.com/open?id=0B4jK5QX65b0ZdDAxbDRBcndvUzA :)

kunall17:
@Sam1301 Thanks for the correct information! :) 
@timabbott fixed the merge conflicts!

kunall17:
@timabbott fixed the merge conflicts! :)

timabbott:
@kunall17 can you post screenshots of the current version?

Also, @Sam1301 can you give this another review?

kunall17:
Sure 
device-2017-01-07-021935

Sam1301:
sure :+1:

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -2002,5 +2014,75 @@ public void onNewMessages(Message[] messages) {
...
+        v.setLayoutParams(snackBarParams);
+        TypedValue tv = new TypedValue();
+        if (getTheme().resolveAttribute(android.R.attr.actionBarSize, tv, true))
+            mToolbarHeightInPx = TypedValue.complexToDimensionPixelSize(tv.data, getResources().getDisplayMetrics());
+    }
+
+    NarrowFilter narrowFilter;
+
+    private void showSnackbarNotification(Message[] messages) {
+        MutedTopics mutedTopics = MutedTopics.get();
```
Sam1301: @kunall17 I played around with the notifications and found that if messages from one particular stream and subject are posted sequentially, the notification shows no ui change, just "1 new message in stream/subject". I am not sure if this is intentional.
Here the messages posted are not quick enough as to come together in a single event and not too slow enough so that  the snackbar from previous notification vanishes. Maybe a check that the snackbar is off the screen can help resolve this.

Sam1301:
@kunall17 I have added a comment.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
kunall17: Hmmm, yeah this is intentional what could be the possible solution be? 
Let the snackbar disappear and display the snackbar again?

Sam1301: maybe dismissing the previous notification snackbar in onNewMessages() can help...

kunall17:
@Sam1301  updated!

Sam1301:
Awesome, the code looks good to me :+1:

kunall17:
@Sam1301 this one is ready to merge right, review full done?

Sam1301:
yep :+1:

kunall17:
@timabbott LGTM, final review by @sam1301 ! :)

kunall17:
@timabbott bump! :)

timabbott:
Merged, thanks @kunall17!

PR #225, opened 2016-11-13
Register app to share all types of files and add attachment uploads feature

### LINKED ISSUE #398: Add attachment uploads feature.
**Actual Behaviour**
Currently, we only support sharing images from outside the app using intent filter.

**Expected Behaviour**
Support attachment uploads to share all types of files from within the app.

Sam1301:
Summary of changes
Fixes #398
* add support for sharing files
To allow sharing of all types of files I have used the following intent filter.
```
<intent-filter>
                <action android:name="android.intent.action.SEND" />
                <category android:name="android.intent.category.DEFAULT" />
                <data android:mimeType="*/*" />
</intent-filter>
```
* add support for attachment uploads
notifications are used to during uploads.
notification
* multiple uploads supported
* implemented a dialog fragment with custom layout for photo and file uploads buttons.
screenshot_20170225-045432
screenshot_20170225-045457
* fixed filepathheper

I haven't been able to achieve desired performance for progress update because retrofit 2 doesn't support progress update during network calls and hence I implemented a hack to work around it. I'll update the pr as soon as I find a better solution for this.

EDIT: Fixed the performance issue.


smarx:

timabbott:
@kunall17 @niftynei can one of you review this?

kunall17:
@timabbott She's about to update this PR with some more enhancements, so i was just waiting for her to make those changes!

kunall17:
We should move the upload icon from the toolbar to the chatBox its not relevant in the toolbar!

Sam1301:
hmm.. I was worried that the chatbox will look quite filled up on small screen devices. I will make the changes :+1:  
Also should we change the upload icon to [attachment icon](https://material.io/icons/#ic_attach_file).. like in the web app?

kunall17:
Yeah that would make more sense!

saketkumar95:
@kunall17 @Sam1301 I think the chatBox will really be quite filled up. We can have the all the sharing things in overFlow menu , just like whatsapp have. What do you think?

Sam1301:
yeah thats a nice idea or maybe we can have a combined button for photo uploads and attachments using a dialog like in the Slack app, @kunall17 what do you think?

saketkumar95:
@Sam1301 Yeah the dialog part would be best for now i guess. :+1:

kunall17:
How about something like this 
https://github.com/oguzbilgener/CircularFloatingActionMenu/raw/master/assets/circularfab.gif

Which can be customized to buttons like here https://github.com/oguzbilgener/CircularFloatingActionMenu/raw/master/assets/customViews.png

We can use this library?

saketkumar95:
@kunall17 Great. But can we just avoid using some more libraries? There is already huge number of libraries which increase the apk size alot.

kunall17:
@saketkumar95 What other options do we have? 
It's better if we can have all the buttons related to chatBox in the chatBox 
And in the future we have one emoji popup as well, so probably we can include that as well in this button!

saketkumar95:
@kunall17 Hmm. Yeah then we can use this library :)

Sam1301:
just want to add this info, that in the Slack app, + button (for photos and upload files) and send button are animated to occupy the same position in the chatbox. If we go with this approach, then I guess the first step would be to have three buttons separately (emoji picker, +, and send) and then at a later milestone add animation to combine + and send. Thoughts?

kunall17:
> at a later milestone add animation to combine + and send. 

I think the send button should be isolated and different from the other buttons, what if someone wants to upload a photo after writing something? 

So we can rather have the + button in the left side and show the emoji, and other stuff!

saketkumar95:
@kunall17 @Sam1301 I have an idea. We can do like this,  suppose when message editText is null we can display attachment icon visible instead of the send button, and when something is written or not null then we can have the send button visible and the attachment icon gone. In the attachment, we can have a option like open camera which is already there and an option of picking the file or image from galary. And in the left side , we can have the emoji picker.  What are your thoughts on this?

saketkumar95:
>  what if someone wants to upload a photo after writing something?

I saw most apps doesn't supports this. But on zulip web we can have this so i guess we need to do this on the app too.  

I'm done with this https://github.com/zulip/zulip-android/issues/248 issue but that time this https://github.com/zulip/zulip-android/pull/228 wasn't merged , so i placed the emoji on left side only.  after the merge , i placed that beside the send button. Now what? Should i use this library https://github.com/oguzbilgener/CircularFloatingActionMenu/ to add emoji and shift photo upload there? 

@Sam1301 I guess we need to work on this together. One cannot wait  to complete this and other to start on this part only.  What say?

timabbott:
I think we have come flexibility for how to do this when prototyping, but (1) we probably don't really need to have the upload/emoji picker items visible when the keyboard isn't open; I feel like either you're in "entry stuff" mode or not and (2) ideally it'd be possible to enter emoji etc. after other text.

Sam1301:
I feel that using circularflowactionmenu library will increase the number of clicks by one while composing messages (which won't even have an upload or emoji). @saketkumar95 i think you can place the emoji picker to the right... beside the send button for now. I think this is a relatively minor issue compared to both the features upload/emoji picker. Hence we can also change it or redesign it easily after both features are merged or one of them is merged :) what say?

kunall17:
I would say either of you send a PR for the extra button on the left and we can get it merge ASAP and then use the button to create emoji/uploader button! 

Also https://github.com/oguzbilgener/CircularFloatingActionMenu/ is not updated recently so maybe we can find some other of the similar task!

Sam1301:
>I would say either of you send a PR for the extra button on the left and we can get it merge ASAP and then use the button to create emoji/uploader button!

ok so a "+" button which opens a dialog with option "Take picture" for now? Just a follow up question, we are gonna merge this change before release?

saketkumar95:
@kunall17 We can use the Horizontal Menu of this library  https://github.com/HarinTrivedi/FABRevealMenu-master

It looks awesome to me. What do you think?

saketkumar95:
@Sam1301 @kunall17 what i think is , we should implement this library and shift the photo upload there and merge it. After that we can separately work on @Sam1301 on attachment and i will be on emoji picker. what say?

if this library looks good , then i can implement this and send the PR ? thoughts?

Sam1301:
The library looks nice! The only thing that i am worried about is currently fab opens the chatbox and keyboard to start typing the message. If we add such complexity to the fab, the user will have to click more to compose a message.

Sam1301:
@kunall17 I implemented the initial work for '+'  button. I can change this if it doesn't look good. Thoughts?

saketkumar95:
@Sam1301 The idea is good. Now we can have emoji picker beside the send button. But don't know maybe it will look like filled up.

Sam1301:
changes:
* the perf is much better for progress update
* resolved merge conflicts

Sam1301:
just resolved merge conflicts

kunall17:
Got an error on first launch (fresh install) 

```
03-03 16:47:17.091 11736-11736/com.zulip.android.dev E/AndroidRuntime: FATAL EXCEPTION: main
                                                                       Process: com.zulip.android.dev, PID: 11736
                                                                       java.lang.RuntimeException: Unable to destroy activity {com.zulip.android.dev/com.zulip.android.activities.ZulipActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.NotificationManager.cancelAll()' on a null object reference
                                                                           at android.app.ActivityThread.performDestroyActivity(ActivityThread.java:3930)
                                                                           at android.app.ActivityThread.handleDestroyActivity(ActivityThread.java:3948)
                                                                           at android.app.ActivityThread.access$1500(ActivityThread.java:154)
                                                                           at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1445)
                                                                           at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                           at android.os.Looper.loop(Looper.java:234)
                                                                           at android.app.ActivityThread.main(ActivityThread.java:5526)
                                                                           at java.lang.reflect.Method.invoke(Native Method)
                                                                           at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                           at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
                                                                        Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.NotificationManager.cancelAll()' on a null object reference
                                                                           at com.zulip.android.activities.ZulipActivity.onDestroy(ZulipActivity.java:2406)
                                                                           at android.app.Activity.performDestroy(Activity.java:6456)
                                                                           at android.app.Instrumentation.callActivityOnDestroy(Instrumentation.java:1143)
                                                                           at android.app.ActivityThread.performDestroyActivity(ActivityThread.java:3917)
                                                                           at android.app.ActivityThread.handleDestroyActivity(ActivityThread.java:3948) 
                                                                           at android.app.ActivityThread.access$1500(ActivityThread.java:154) 
                                                                           at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1445) 
                                                                           at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                                           at android.os.Looper.loop(Looper.java:234) 
                                                                           at android.app.ActivityThread.main(ActivityThread.java:5526) 
                                                                           at java.lang.reflect.Method.invoke(Native Method) 
                                                                           at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726) 
                                                                           at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616) 
```

kunall17:
This is a really small thing, but don't you think the dialog looks a bit down and not centered because of absence of toolbar?  
I can imagine on a device like one plus 3 which allows removing this bar can cause offset a lot, perhaps (or maybe i'm thinking too much)

Sam1301:
yeah, the dialog is centered vertically below the status bar by default. But I couldn't see it to be centered below toolbar? I enabled layout boundaries on my device to see this.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -563,22 +575,44 @@ public Cursor runQuery(CharSequence charSequence) {
...
         this.getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_STATE_HIDDEN);
     }
 
+    public void showListDialog() {
+        // Create an instance of the dialog fragment and show it
+        DialogFragment dialog = new ListDialog();
+        dialog.show(getFragmentManager(), "ListDialogFragment");
+    }
+
+    // The dialog fragment receives a reference to this Activity through the
```
kunall17: Put these comments in javadoc comments 
/*
*
*/

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -838,18 +877,49 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
             // activity transition animation
             ActivityTransitionAnim.transition(ZulipActivity.this);
+        } else if (requestCode == REQUEST_PICK_FILE && resultCode == RESULT_OK) {
+            List<Uri> fileUris = new ArrayList<>();
+            if (data.getData() != null) {
+                fileUris.add(data.getData());
+            }
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
kunall17: Is this fine, generally there is a if condition for the versions below 16 (JELLY BEAN) ? 
If yes add a comment here!

kunall17:
Also it would be better to squash 'Reformat ZulipActivity'  into there respective commits! (a232dd403ad15f49bb7158dc3cc417144b82e1d7)

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -124,30 +133,29 @@
...
  */
 public class ZulipActivity extends BaseActivity implements
-        MessageListFragment.Listener, NarrowListener, SwipeRemoveLinearLayout.leftToRightSwipeListener {
+        MessageListFragment.Listener, NarrowListener, SwipeRemoveLinearLayout.leftToRightSwipeListener,
+        ListDialog.ListDialogListener {
 
     private static final String NARROW = "narrow";
     private static final String PARAMS = "params";
     //At these many letters the emoji/person hint will not show now on
     private static final int MAX_THRESOLD_EMOJI_HINT = 5;
```
kunall17: I meant these comments 
```
private static final int HIDE_FAB_AFTER_SEC = 5;
private static final int REQUEST_PICK_FILE = 3;
```

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
Sam1301: added a comment, intent.getClipdata was added api 16

Sam1301:
>Have some description in the notification for which stream or private it being uploaded to?

added title description in start notifications

kunall17:
For the button to notification 

You can do something like this in the setNotification() method!
```
        Intent buttonIntent = new Intent(this, GcmBroadcastReceiver.class);
        buttonIntent.setAction(ACTION_HERE);
        buttonIntent.setType("TYPE_HERE");
        PendingIntent pendingIntentYes = PendingIntent.getBroadcast(this, CONST_HERE, buttonIntent, PendingIntent.FLAG_UPDATE_CURRENT);
        mBuilder.addAction(R.drawable.ic_cancel_black_24dp, "Dismiss", pendingIntentYes);
```


And as we know the upload's gonna stop if the app is destroyed hence we can use the ZulipApp.get().getActivity() [or something like this name method] to get the current activity and do the UI stuff here if requires, I think we just need to dismiss the notification and show a Toast that it's cancelled! 

Make sure to add some checks in the GCMBroadcastReceiver class onRecieve method

I used the GcmBroadcastReceiver because it's already registered and being used, hence we can re-use this, we might want to rename this file! 

Also the name still isn't right as shown in the GIF!

Sam1301:
thanks for providing a better approach :+1: , did exactly as suggested + used a hashmap for notificationId, Call objects.

Sam1301:
@kunall17 addressed all suggestions, can you please take a look?

app/src/main/java/com/zulip/android/gcm/GcmBroadcastReceiver.java
```diff
@@ -22,13 +29,30 @@
...
     private static final String TAG = "GCM";
-
+    public static final String CANCEL = "Cancel";
     public static String getGCMReceiverAction(Context appContext) {
         return appContext.getPackageName() + ".PushMessage.BROADCAST";
     }
 
     @Override
     public void onReceive(Context context, Intent intent) {
+        // handle cancel notification action on uploads
```
kunall17: Declaring the hashMap private and calling a function from here to zulipActivity would be better! 
Also add a check if `ZulipApp.get().getZulipActivity().hashMap;` returns null, as suppose if the app gets destroyed and notification is still hanging there then there would be crash!

app/src/main/java/com/zulip/android/gcm/GcmBroadcastReceiver.java
```diff
@@ -22,13 +29,30 @@
 public class GcmBroadcastReceiver extends BroadcastReceiver {
 
     private static final String TAG = "GCM";
-
+    public static final String CANCEL = "Cancel";
```
kunall17: This to be sent to Constants.java and `CANCEL = "CANCEL"; `

app/src/main/java/com/zulip/android/gcm/GcmBroadcastReceiver.java
Sam1301: yes, thats why i have surrounded the block with a try catch

Sam1301:
@kunall17 updated the branch (travis passed).

kunall17:
I can still see this name for the file I sent, any reasons?

Sam1301:
@kunall17 I couldn't reproduce this error, i am updating the file name here https://github.com/zulip/zulip-android/pull/225/files#diff-93e07d12c8f17453899c8a8cceb659d2R1060. Is this true just for a single file for you?

Sam1301:
@kunall17 can you please have a look?

kunall17:
- When the upload starts successfully display a toast, that the file is being uploaded
- If I try to upload a file from the SD card, it fails shows `Invalid File` after your recent changes

kunall17:
The error occurs for the `FileUtils.getPath` 
if the uri of the file (from sd card) is 
`content://com.android.externalstorage.documents/document/5865-3F42%3AMovies%2FScreenRecords%2FScreenRecord_2016-12-27-23-14-59.mp4`
And the path which FileUtils.getPath returns is `/storage/emulated/0` for the `filePath` String

Which is wrong, because it uses Environment.getExternalStorageDirectory() which returns different values for different manufactures, for example mine is sony and this value returns `/storage/emulated/0` while the correct path is `/storage/5865-3F42/ `
Hence this problem occurs! 
Have a look here http://stackoverflow.com/a/23124547

Sam1301:
@kunall17 updated the  branch... tested for samsung sd card uploads.

kunall17:
I hate to say this, but it still doesn't works
https://github.com/Sam1301/zulip-android/blob/e33afd9a4208fb699a6233f9b0e3ad8eb80f6c84/app/src/main/java/com/zulip/android/util/FileUtils.java#L280

Here you have explicitly compared the "primary" with the type string, in my sony phone it's value was 5865-3F42 so the function returned null! :(

And you have modified the FileUtils file which is a licensed file, and according to [Apache version 2.0](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)) the changes must be stated and in the file also its mentioned `distributed on an "AS IS" BASIS,` so its better not to edit this file!

Sam1301:
@kunall17 done :+1: , as we discovered it was all due to android fragmentation :P

kunall17:
Awesome work, merged thanks @Sam1301 👍

Issue #227, opened 2016-11-16
"Failed to fetch Backends" 


jmandel:
On the beta channel, when I try to enter a Zulip server URL (e.g. https://zulip.tabbott.net or https://chat.fhir.org) the value "/api/" is automatically appended, and then I get the error:

"Failed to fetch Backends" 

Is this a known issue?

timabbott:
No; are you by any chance typing `http://` rather than `https://`?

jmandel:
No, though I did try that for fun, and I get a security popup asking whether I really wanted to use HTTP (and saying "yes" led to the same error).

timabbott:
Interesting.  @niftynei any idea what in the beta could have caused this?  It's not a big diff from the production release...

jmandel:
BTW is there an easy way for me to test the production release (move off the beta channel)? Alternatively, can anyone report whether they can reproduce this behavior (in beta or prod)?

timabbott:
I can't reproduce with 1.1.1 (the beta)

kunall17:
Production release is at this branch https://github.com/zulip/zulip-android/tree/release-candidate
I also use the beta app and it works fine!

jmandel:
It appears that the android client relies on `/api/v1/get_auth_backends` (introduced [here](https://github.com/zulip/zulip/commit/a36e5d4987d9ac8128c57dab4e5c42ded175f304)), and https://chat.fhir.org is running a version of Zulip old enough not to have one.

But that's not the whole issue, because I get the same error message for https://zulip.tabbott.net, which *does* have one.

niftynei:
thanks for the report. fwiw the error message you're receiving isn't very precise.  a failure to get backends may happen for a variety of reasons, including a network failure.  can you verify that you're able to reach the website on a different app?

jmandel:
To confirm, yes, I can reach these sites through the browser just fine. Is that what you mean,  @niftynei?

timabbott:
Does https://chat.zulip.org work for you as an Android auth point?  I wonder if the old zulip.tabbott.net name is what's busted.

jmandel:
No, https://chat.zulip.org/ also fails, with the same error. For what it's
worth, everything was working just fine until I signed out and tried to
sign back in to demonstrate the process for a colleague..

timabbott:
Oh, that detail is interesting.  I wonder if it's some sort of caching bug...

jmandel:
[...]

PR #229, opened 2016-11-27
Fix bug#227

Sam1301:
This pr addresses #227 
The server url first entered (when the app is started) was being used every time, regardless of what the user typed in next. Therefore, I added a new function `setZulipServices()` in `ZulipApp.java` and used it to set `zulipServices` to `null` before calling `getServices()` in `LoginActivity`. This will cause new `zulipServices` object creation **only during login**.
I have also hidden `/api` from the user and removed any leading/trailing spaces in `serverUrl` entered by the user.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/LoginActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/LoginActivity.java
@@ -151,6 +151,10 @@ public class LoginActivity extends BaseActivity implements View.OnClickListener,
 
     private void checkForError() {
         String serverURL = serverIn.getText().toString();
+
+        // trim leading or trailing white spaces in Url
+        serverURL = serverURL.trim();
+
         int errorMessage = R.string.invalid_server_domain;
         String httpScheme = (BuildConfig.DEBUG) ? "http" : "https";
 
@@ -184,6 +188,11 @@ public class LoginActivity extends BaseActivity implements View.OnClickListener,
     }
 
     private void showBackends(String httpScheme, String serverURL) {
+        // if server url does not end with "/", then append it
+        if (!serverURL.endsWith("/")) {
+            serverURL = serverURL + "/";
+        }
+
         Uri serverUri = Uri.parse(serverURL);
 
         serverUri = serverUri.buildUpon().scheme(httpScheme).build();
@@ -197,6 +206,9 @@ public class LoginActivity extends BaseActivity implements View.OnClickListener,
         mServerEditText.setText(serverUri.toString());
         mServerEditText.setEnabled(false);
         ((ZulipApp) getApplication()).setServerURL(serverUri.toString());
+
+        // create new zulipServices object every time by setting it to null
+        getApp().setZulipServices(null);
 
         getServices()
                 .getAuthBackends()
--- a/app/src/main/java/com/zulip/android/activities/LoginActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/LoginActivity.java
@@ -197,14 +197,17 @@ public class LoginActivity extends BaseActivity implements View.OnClickListener,
 
         serverUri = serverUri.buildUpon().scheme(httpScheme).build();
 
-        // if does not begin with "api.zulip.com" and if the path is empty, use "/api" as first segment in the path
-        List<String> paths = serverUri.getPathSegments();
-        if (!serverUri.getHost().startsWith("api.") && paths.isEmpty()) {
-            serverUri = serverUri.buildUpon().appendEncodedPath("api/").build();
-        }
+        // display server url with http scheme used
         serverIn.setText(serverUri.toString());
         mServerEditText.setText(serverUri.toString());
         mServerEditText.setEnabled(false);
+
+        // if server url does not end with "api/" or if the path is empty, use "/api" as last segment in the path
+        List<String> paths = serverUri.getPathSegments();
+        if (paths.isEmpty() || !paths.get(paths.size() - 1).equals("api")) {
+            serverUri = serverUri.buildUpon().appendEncodedPath("api/").build();
+        }
+
         ((ZulipApp) getApplication()).setServerURL(serverUri.toString());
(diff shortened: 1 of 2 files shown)


smarx:

timabbott:
@kunall17 @niftynei if one of you can review this, I can take care of uploading a new beta based on it for @jmandel to test.

Sam1301:
Updated the pr to take care of some more end cases. The app crashes if the server url doesn't end with `/`. Also, the `api/` should be added at the end of server url if the paths segment is empty or the last path provided by the user is not `api/`.

app/src/main/java/com/zulip/android/activities/LoginActivity.java
```diff
@@ -184,20 +188,31 @@ public void onSaveInstanceState(Bundle savedInstanceState){
...
         Uri serverUri = Uri.parse(serverURL);
 
         serverUri = serverUri.buildUpon().scheme(httpScheme).build();
 
-        // if does not begin with "api.zulip.com" and if the path is empty, use "/api" as first segment in the path
-        List<String> paths = serverUri.getPathSegments();
-        if (!serverUri.getHost().startsWith("api.") && paths.isEmpty()) {
-            serverUri = serverUri.buildUpon().appendEncodedPath("api/").build();
-        }
+        // display server url with http scheme used
```
kunall17: Can you split this another commit as it has nothing to do with the issue! 
Btw why do you want to hide the /api  being appended in the server URL? 
I think it was ok to display the complete url!

app/src/main/java/com/zulip/android/activities/LoginActivity.java
```diff
@@ -184,20 +188,31 @@ public void onSaveInstanceState(Bundle savedInstanceState){
...
+        // if server url does not end with "api/" or if the path is empty, use "/api" as last segment in the path
+        List<String> paths = serverUri.getPathSegments();
+        if (paths.isEmpty() || !paths.get(paths.size() - 1).equals("api")) {
+            serverUri = serverUri.buildUpon().appendEncodedPath("api/").build();
+        }
+
         ((ZulipApp) getApplication()).setServerURL(serverUri.toString());
 
+        // create new zulipServices object every time by setting it to null
+        getApp().setZulipServices(null);
```
kunall17: I think we can move this line to onError and onFailure methods, as if the connection is successful why to recreate ZulipServices object again?

app/src/main/java/com/zulip/android/activities/LoginActivity.java
Sam1301: Sure :+1:  
As mentioned by @timabbott here https://github.com/zulip/zulip-android/issues/227, we should hide 'api/' from user as old api.zulip.com doesn't exist anymore.

app/src/main/java/com/zulip/android/activities/LoginActivity.java
kunall17: Yeah that makes sense, I didn't gave a thought about logout.
If you would set the ZulipServices null after logging out it would work, as the retrofit baseURL is what making the difference.
I am just trying to avoid the re initialization of the ZulipServices after logging in.

app/src/main/java/com/zulip/android/activities/LoginActivity.java
kunall17: :+1:

app/src/main/java/com/zulip/android/activities/LoginActivity.java
Sam1301: @kunall17 Can you help me with this, I cannot think about the case when `showBackends()` will be called after logging in which would lead to the unnecessary re initialization of `zulipServices`.

kunall17: Even after successful login the ZulipServices will be null, now when AsyncGetEvents calls getZulipServices  here https://github.com/zulip/zulip-android/blob/c9855cf2b4f531b325e65fe324a362be75396553/app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java#L107

Those objects will be re initialized!

kunall17: Infact getServices is being called in the next line, didn't noticed this one! 
https://github.com/Sam1301/zulip-android/blob/d8eb3d7bd001482e6873109ac92ed85b0dc15817/app/src/main/java/com/zulip/android/activities/LoginActivity.java#L216

Sam1301: Thanks @kunall17 for clearing this out.
I am not sure if I got your point. I just tried to debug the app and couldn't find `zulipServices` being recreated. After `zulipServices` object is initialized(during a successful login), the if condition `if (zulipServices == null)` in getZulipServices() prevents it from being initialized again.

kunall17: Ok, ya the code is fine, got confused somewhere.. :+1:

kunall17:
@timabbott lgtm .. :+1:

app/src/main/java/com/zulip/android/activities/LoginActivity.java
Sam1301: Thanks for reviewing this @kunall17 :+1:

timabbott:
@Sam1301 can you clean up the commit messages to match the standard Zulip style?  

Check out the Zulip commit message guidelines for more details: https://zulip.readthedocs.io/en/latest/version-control.html#commit-messages

Sam1301:
Thanks for the guide, updated the PR :+1:

timabbott:
@Sam1301 I just went ahead and fixed this for these commits while merging this, but you need a blank line between the first line of your commit messages and the later lines.  

Merged, thanks @Sam1301!

Sam1301:
Thanks! I'll take care of this from next time.

PR #234, opened 2016-12-03
Fixes #233

### LINKED ISSUE #233: Quality Assurance: App crashes when pinging users with '@'
The Android app crashes when I try to ping a user in a message with '@'. More specifically, it crashes when typing the first letter after the '@'.

Sam1301:
This pr addresses #233 
The existing code relies on column index to distinguish between emoji or person cursor in `combinedAdapter`. As a good database practice, _we should never rely on the defined order of fields in any database_
Hence, I have used total number of columns to see which cursor is being used.
Somehow, this error occurs only for devices running on android versions 5.0 or less. I have test my changes on versions 4.x, 5.x and 6.x

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -497,8 +497,9 @@ public class ZulipActivity extends BaseActivity implements
                 int index = cursor.getColumnIndex(Emoji.NAME_FIELD);
                 String name = cursor.getString(index);
                 String currText = messageEt.getText().toString();
-                int last = (cursor.getColumnIndex(Emoji.NAME_FIELD) == 6) ? currText.lastIndexOf("@") : currText.lastIndexOf(":");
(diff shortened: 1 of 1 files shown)


smarx:

kunall17:
Thanks for the work @Sam1301 
Can you work on the bug of unread counts?

Sam1301:
@kunall17 Sure. Just to be clear, can you give me a reference to this bug. I can see 2 issues with unread counts problem.

kunall17:
@Sam1301 
No child's (topics) are generated for the streamsDrawer expandableListView
The code line (SQL Query) which is causing the problems https://github.com/zulip/zulip-android/blob/946a329c5483385c0bbfeb140732bdff8ca92c40/app/src/main/java/com/zulip/android/activities/ExpandableStreamDrawerAdapter.java#L35

The getChildrenCursor is called whenever you click on the group, but the SQL query here is going wrong hence it returns 0 currently therefore no childs (topics) are generated, 
In this SQL query I also calculated the unread counts in a separate column!

timabbott:
Merged as db066ac5f61bccffa7359c8b88ef23f8ff3326c3 (I added a more descriptive commit message).  Thanks @Sam1301!

PR #241, opened 2016-12-14
Re-formatted whole project using the default code formatting

kunall17:
This basically re-formats and re-arranges the code lines according to android studio default code formatting,
For the new incoming PR's some have un-formatted code and if they use the formatting code command (To format the code they have written) it would format the whole current file therefore the commit includes changes in the code which they haven't touched,
So after this commit it would be easy for them to format code rather than selecting their changes and formatting them to avoid inclusion of git changes of other formatted code.
Hope this makes sense! :)

### DIFF
--- a/app/src/androidTest/java/com/zulip/android/activities/BaseTest.java
+++ b/app/src/androidTest/java/com/zulip/android/activities/BaseTest.java
@@ -37,12 +37,15 @@ import static org.junit.Assert.assertNull;
 @FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class BaseTest {
 
+    public static final String SERVER_URL = "www.local.test.com";
     private static final String PASSWORD_TEST = "yourpasswordhere";
     private static final String EMAIL_TEST = "iago@zulip.com";
-    public static final String SERVER_URL = "www.local.test.com";
-
     @Rule
     public ActivityTestRule<ZulipActivity> mActivityTestRule = new ActivityTestRule<>(ZulipActivity.class);
+
+    public static Matcher<Root> isToast() {
+        return new ToastMatcher();
+    }
 
     @Before
     public void setUp() {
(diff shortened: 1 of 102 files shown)


smarx:

timabbott:
@niftynei @Sam1301  what are your thoughts on this reformatting?  

If folks are on board with doing it I'm happy to merge, though I think we should clear the open PRs first since this will create a lot of merge conflicts.

Sam1301:
I really like this idea. There are a lot of places in the project where code factoring is needed. While submitting new prs, people generally use built in shortcuts in Android studio for code factoring in android. But this also affects other places where changes were not made and hence messes up the git history. So, once this is merged we can use these shortcuts 👍

timabbott:
OK.  I think we should do a coordinated effort to try to merge the useful open PRs and then do this, so as to avoid needing to merge the automated formatting changes with actual code changes :)

kunall17:
Yeah, we should merge the rest of active PR's first! 
@Sam1301 can you review the #217 ?

Sam1301:
Sure, will do this today :+1:

timabbott:
I'm not sure that we're actually on track to have fewer open PRs; @kunall17 @saketkumar95 @Sam1301 should we change our strategy for this?

kunall17:
I think we should merge this anyway, waiting for no open PR's would be a pain and practically very difficult to achieve.
If yes i would fix the merge conflicts we are getting!

timabbott:
@saketkumar95 @Sam1301 what do you think?

saketkumar95:
@timabbott Yeah @kunall17  is right . We should merge this soon because i guess there will be more open PRs in future and will achieving this after that will be a lot difficult.

timabbott:
OK, sounds good, let's just get this in.

kunall17:
Updated the branch! :)

timabbott:
Merged, thanks @kunall17!  We should probably post comments on all the open PRs letting folks know we did this and they'll need to rebase.

Issue #246, opened 2016-12-15
Don't open the keyboard immediately on email/password login page


timabbott:
Since on Android, Google auth is usually going to be what folks want, we should probably not open the keyboard immediately (since it sorta obscures the Google auth button) and require users to click into the text boxes to start typing.  What do you think @kunall17?

saketkumar95:
@timabbott we can do so just by implementing callback methods from the KeyEvent.Callback interface, such as onKeyDown() and onKeyMultiple().

kunall17:
Yeah, we should do that! 
@saketkumar95 can you test this and send a PR for this one? 
 The manifest edit is the better approach!

saketkumar95:
Yes. Sure . i will send the PR by morning today.

PR #250, opened 2016-12-16
fixes: #246 

saketkumar95:
fixes: [#246](https://github.com/zulip/zulip-android/issues/246) 

@kunall17 @timabbott anyone of you do review this.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -48,7 +48,7 @@ android {
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.txt'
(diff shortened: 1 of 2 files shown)


smarx:

@saketkumar95, it looks like you've already signed the Dropbox CLA. Thanks!

kunall17:
@saketkumar95 As we discussed to use manifest approach, but this won't do because the first screen asks for the server URL there we want that the keyboard should automatically be displayed, and after logging in there keyboard should not be displayed!  
Also\
- Fix the commit history (there are two commits right now) 
- Reword the commit message to something (Fixes #246: Disable auto showing up the soft keyboard in login activity)

Issue #249, opened 2016-12-16
Register new account


maxlinux2000:
It would be nice if the user can create a new account from zulip-android

timabbott:
Agreed.  It's kinda expensive to maintain 2 copies of the registration flow; we might just want to provide a nicely integrated way to get to that flow from the Android app; I guess this would need to be a link from the page after you choose your server.

kunall17:
@timabbott for this one, we could just add a link after entering the server and open the browser in a chrome custom tabs? 
Now after registration, user can manually enter credentials.
A better (but far more complex) would be deeplinking the android project which would send an API key to the android client after registration, and use this key to login?

timabbott:
Yep.  Let's do the v1 with the browser and then we can try to figure out how to deeplink as a potential followup.

vishwesh3:
I would like to work on this.

kunall17:
@vishwesh3 sure go ahead, with the first step!

PR #253, opened 2016-12-19
Added register link in login page

vishwesh3:
videotogif_2016 12 19_11 07 39
issue #249

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/LoginActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/LoginActivity.java
@@ -444,6 +444,11 @@ public class LoginActivity extends BaseActivity implements View.OnClickListener,
                     }
                 });
                 asyncDevGetEmails.execute();
+                break;
+            case R.id.register:
+                Intent intent = new Intent(Intent.ACTION_VIEW,Uri.parse(serverIn.getText().toString()+"register"));
+                startActivity(intent);
+                break;
             default:
                 break;
         }
--- a/app/src/main/java/com/zulip/android/activities/LoginActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/LoginActivity.java
@@ -5,7 +5,9 @@ import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.IntentSender.SendIntentException;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
+import android.support.customtabs.CustomTabsIntent;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.Toolbar;
 import android.util.Log;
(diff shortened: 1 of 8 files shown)


smarx:

@vishwesh3, thanks for the pull request! It looks like you haven't yet signed the Dropbox CLA. Please [sign it here](https://opensource.dropbox.com/cla/).

app/src/main/res/values/strings.xml
```diff
@@ -94,4 +94,6 @@
     <string name="cannot_find_image">Could not find the image</string>
     <string name="failed_to_upload">Failed to upload</string>
     <string name="invalid_image">Unable to load that image</string>
+    <string name="register_button">Register</string>
+    <string name="default_server_url">https://api.zulip.com/</string>
```
Sam1301: I guess you are not using `default_server_url`, maybe you forgot to remove it.

app/src/main/res/layout-v21/login.xml
```diff
@@ -140,6 +140,18 @@
             android:visibility="gone" />
 
         <TextView
+            android:id="@+id/register"
```
Sam1301: Nice! How about improving it further by using styles to club common properties of Another server, Register and legal info TextViews?

vishwesh3: Which style should I use? Any suggestion.

app/src/main/res/values/strings.xml
vishwesh3: Sorry I forgot to remove that. Now I have removed. Thanks

app/src/main/res/layout-v21/login.xml
```diff
@@ -140,6 +140,18 @@
...
+            android:layout_width="wrap_content"
```
saketkumar95: @vishwesh3 You can style it in the same way and sequence , Like the legal info and another server textViews. :)

kunall17:
Can you use Chrome custom tabs for the registration process https://developer.chrome.com/multidevice/android/customtabs
It would be much better, and can provide better flow once we do changes from the server side as well, like using the registered account to auto sign in after registration!

app/src/main/java/com/zulip/android/activities/LoginActivity.java
```diff
@@ -444,6 +444,11 @@ public void onTaskFailure(String result) {
                     }
                 });
                 asyncDevGetEmails.execute();
+                break;
+            case R.id.register:
+                Intent intent = new Intent(Intent.ACTION_VIEW,Uri.parse(serverIn.getText().toString()+"register"));
```
kunall17: And also please don't forget to format the code!

app/src/main/res/layout-v21/login.xml
Sam1301: Its generally a best practice to group common attributes values in xml tags in styles https://developer.android.com/guide/topics/ui/themes.html. Since Register, Another Server and Legal Info textviews are pretty similar we can group their common xml attributes in a style.

vishwesh3: Done https://github.com/zulip/zulip-android/pull/253/commits/818259ad834b9a52cfc74819509c040ec764b2b0 Thanks

vishwesh3:
@kunall17 customtabs requires minSdkversion 15 and our current is 13 should I change it to 15?

kunall17:
Well increasing the minSdkVersion doesn't seem to be a great solution, We should just override the library to be used, as the probability of using registration with API 13,14 would be very less  http://stackoverflow.com/questions/27335889/how-do-i-ignore-minsdkversion-of-library-in-android-studio

And just use the intents if the device is running API 13,14!

vishwesh3:
@kunall17 Implemented CustomTabs https://github.com/zulip/zulip-android/pull/253/commits/c77a2278081ec7d0a480adab99110a3a72ab33d9 it is much better than old tradition of opening default browser, faster and convenient and doesn't break the flow of the app. Thanks

kunall17:
@vishwesh3 nice work.. :+1: :)
The code looks good to me! 

Btw, consider installing the development server not good to rely upon zulip.tabbott.net for testing purposes!
 
@timabbott LGTM!

vishwesh3:
Thanks @kunall17 I have successfully installed development server for testing. :smile:

timabbott:
Merged, thanks @vishwesh3  and @kunall17!!

smarx:
				
@vishwesh3, thanks for signing the CLA!

PR #254, opened 2016-12-19
Fixes the expandableListView in the streams drawer

kunall17:
This was because in the messages table stream column and the read column have null values which should not be the case, I have uploaded the database of the android app which is dumped after the first login to the app, if any one wants to tinker with the queries (Maybe the SQL queries can be made better)
[databases.zip](https://github.com/zulip/zulip-android/files/661897/databases.zip)

Previously the queries were dependent on the id's now they are being compared by names!

2e4ce93 is basically removing 0fea6fbc525390f9a50ab476b7126f9203fb3dee commit, not sure though if this is the best method to do this!

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ExpandableStreamDrawerAdapter.java
+++ b/app/src/main/java/com/zulip/android/activities/ExpandableStreamDrawerAdapter.java
@@ -32,9 +32,9 @@ public class ExpandableStreamDrawerAdapter extends SimpleCursorTreeAdapter {
     public Cursor getChildrenCursor(Cursor groupCursor) {
         int pointer = zulipApp.getPointer();
         List<String[]> results = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


smarx:

Sam1301:
I tried this, after reading messages in homeview, the unread counts updates in streams drawer but after I tap on anyone of the streams, and then check stream drawer unread count, it increases for that stream.

timabbott:
@kunall17 have you had a chance to look at working on this further?

kunall17:
@timabbott The PR does its job well, but there is something else playing with the read values! 
Sam1301 is looking for it! 
And after merging https://github.com/zulip/zulip-android/pull/283 the problem should be fixed, just we would have to revert the commit 0fea6fb

kunall17:
superseded by #283

Issue #264, opened 2016-12-26
"Today" option in menu


vishwesh3:
"Today" option in menu is not required if **Today Message** page is already opened. We can either remove that or add other filter options like **Yestarday** etc .

![videotogif_2016 12 27_00 58 57](https://cloud.githubusercontent.com/assets/18511177/21485927/5d013dd4-cbd1-11e6-9d56-50f6ddd4fdb9.gif)

@timabbott @kunall17 @Sam1301 Can I work on this?

kunall17:
Maybe a option like One day before, and let it go in a loop like if we go Today -> One day before -> One day before -> One day before
It would be an easy task, you need to tinker with the NarrowToday filter!

vishwesh3:
Thanks @kunall17 I will send PR soon.

PR #276, opened 2017-01-02
Fixes: #264 Implemented One day before

vishwesh3:
Addressing issue #264 
Now at home **Today** option is available in menu and then **One day Before** in loop. 

videotogif_2017 01 02_21 22 52


smarx:

kunall17:
Looks nice! Just if there are no messages to display, the screen  'No messages to display' should be shown!

vishwesh3:
@kunall17 Done.
videotogif_2017 01 03_20 03 09

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -420,6 +420,20 @@ public void onMessages(Message[] messages, LoadPosition pos,
...
+            showRecyclerView();
+    }
+
+    /**
+     * hides TextView with no message
+     * show recyclerView
+     */
+    private void showRecyclerView() {
+        recyclerView.setVisibility(View.VISIBLE);
+        emptyTextView.setVisibility(View.VISIBLE);
```
kunall17: emptyTextView can be made View.GONE?

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -524,6 +529,34 @@ public Cursor runQuery(CharSequence charSequence) {
...
+    private void handleOnFragmentChange() {
+        getSupportFragmentManager().addOnBackStackChangedListener(new FragmentManager.OnBackStackChangedListener() {
+            @Override
+            public void onBackStackChanged() {
+                FragmentManager fragmentManager = getSupportFragmentManager();
+                Fragment fragment = fragmentManager.findFragmentById(R.id.list_fragment_container);
+                if (fragment instanceof MessageListFragment)
+                {
+                    String tag = fragment.getTag();
+                    if (menu==null)
```
kunall17: Please format the code in this block!

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -524,6 +529,34 @@ public Cursor runQuery(CharSequence charSequence) {
...
+                        return;
+                    if (tag.equals(getString(R.string.home_frament_tag)))
+                    {
+                        calendar = Calendar.getInstance();
+                        menu.getItem(2).getSubMenu().getItem(0).setTitle(R.string.menu_today);
+                    }else if (tag.equals(getString(R.string.today_fragment_tag)))
```
kunall17: Rather than saving TAGS for fragments in the Strings.xml it would be better to create file in the utils folder for constants.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -524,6 +529,34 @@ public Cursor runQuery(CharSequence charSequence) {
...
+        calendar = Calendar.getInstance();
+    }
+
+    private void handleOnFragmentChange() {
+        getSupportFragmentManager().addOnBackStackChangedListener(new FragmentManager.OnBackStackChangedListener() {
+            @Override
+            public void onBackStackChanged() {
+                FragmentManager fragmentManager = getSupportFragmentManager();
+                Fragment fragment = fragmentManager.findFragmentById(R.id.list_fragment_container);
+                if (fragment instanceof MessageListFragment)
```
kunall17: This `if (fragment instanceof MessageListFragment)` is a nice approach generally but in our project it is MessageListFragment only so this can be removed?

app/src/main/java/com/zulip/android/filters/NarrowFilterToday.java
```diff
@@ -30,6 +33,10 @@ public NarrowFilterToday createFromParcel(Parcel in) {
     public NarrowFilterToday() {
     }
 
+    public NarrowFilterToday(Date date) {
```
kunall17: It would be better to rename this file as after using this constructor its the narrow would not be for today it would be for some other day!

app/src/main/java/com/zulip/android/filters/NarrowFilterToday.java
```diff
@@ -12,10 +12,13 @@
 import org.json.JSONException;
 
 import java.sql.SQLException;
+import java.text.SimpleDateFormat;
 import java.util.Date;
 
 public class NarrowFilterToday implements NarrowFilter {
 
+    private Date date =new Date();
```
kunall17: Code formatting required here!

kunall17:
@vishwesh3 really love this feature :) 
Also you depend upon tags for the fragment rather you should use if (narrowedList==null) for homeView and if (narrowedList!=null) and if ( narrowedList.filter instanceOf NarrowFilterToday) then it is using a filter of narrowToday!

vishwesh3:
@kunall17 Done with the changes.

vishwesh3:
I was thinking to add one more item in the menu **Enter Date**, on selecting it we will ask user to pick Date from DatePicker via Dialog and then Narrow to that date. For example if user want's to see messages of 15 days back it is difficult to go **One day before** 15 times. In such cases **Enter Date**  option will save time and enhance user experience. 

@timabbott @kunall17 What do you think?

kunall17:
@vishwesh3 yeah that would be nice, include this in a menuItem and another commit!

vishwesh3:
Added **Enter Date** in the menu.

videotogif_2017 01 08_14 12 44

kunall17:
This looks awesome, thanks for the work @vishwesh3 ! 👍 
@timabbott LGTM!

timabbott:
Merged, thanks @vishwesh3!!

Issue #145, opened 2016-08-22
On Narrow, Keep Narrowed Conversation in View


niftynei:
Branching from a suggestion on #139.

When narrowing, we should ensure the narrowing is done in a way where the message that you touched to initiate the narrow on stays in view (rather than jolting around). See for notes on how this works in the webapp.

http://zulip.readthedocs.io/en/latest/pointer.html

timabbott:
Yeah this would be super cool.  I'd recommend implementing it in the way the webapp says it would like to be implemented, not the way it currently works.  That may involve some Zulip backend changes -- I'm happy to help make those :)

Sam1301:
I'd like to work on this issue :)

PR #279, opened 2017-01-05
On narrowing, keep messages in streams and private messages in view.

Sam1301:
This pr addresses #145 
I have changed the way we are fetching messages when a message tile is clicked. Now, the message being clicked is used as an anchor while fetching and kept in view on narrowing.


smarx:

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1645,6 +1659,12 @@ public void onNarrow(NarrowFilter narrowFilter) {
         doNarrow(narrowFilter);
     }
 
+
+    public void onNarrow(NarrowFilter narrowFilter, int messageId) {
+        // TODO: check if already narrowed to this particular stream/subject
```
kunall17: What about this TODO?

kunall17:
This one is a biggie but important as well, so what are your thoughts on this one @vishwesh3 @saketkumar95?
Can you guys give a quick review?

vishwesh3:
sure

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
Sam1301: done, forgot to implement it :+1:

Sam1301:
Thanks @vishwesh3, made the changes.

Sam1301:
update: added support for narrowing to last message read in a particular stream when a stream is selected from the streams drawer.

kunall17:
I tested the branch the but the commit f83d070 doesn't works, even if the messages are in the cache and read they aren't scrolled to the last read message! 
ezgif com-20435f322b

Sam1301:
>Now, the message being clicked is used as an anchor while fetching and kept in view on narrowing.

I guess thats because this was a narrow on message click, but not from a stream in the stream drawer.

kunall17:
Ok, didn't read the code before! 
This should be supported in these messageHeaderParent and narrow from the menu on long press on message! 

Works well from the streams drawer 👍

Sam1301:
> This should be supported in these messageHeaderParent and narrow from the menu on long press on message!

sorry I didn't get that, can you please elaborate more on this? I have added code to support narrow to a message clicked keeping the message in view. This is done for messageHeaderParent. Are you suggesting to do the same for long press menu on message?

kunall17:
These changes https://github.com/zulip/zulip-android/pull/279/files#diff-93e07d12c8f17453899c8a8cceb659d2L1157 should be here as well https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L130

Sam1301:
hmm.. Is the reference correct ? I have changed https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L130 to https://github.com/zulip/zulip-android/pull/279/files#diff-6727cdd29f9d849bb181415059bd3279L130 to account for message in view on click.

kunall17:
Nope this interface is implemented in ZulipActivity and this is the function https://github.com/zulip/zulip-android/blob/1229ae6afef8443f10d119edd6bf4ade1924d0f3/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L1798

kunall17:
When I tried on my emulator and dev env it worked but when on zulip.tabbott.net it didn't load older messages! 

By the logs i saw 

> 01-31 22:42:48.777 23876-23876/com.zulip.android.dev I/onMessage: Already have 125414
> 01-31 22:42:48.777 23876-23876/com.zulip.android.dev I/onMessage: Already have 125415

Messages like these in large number! 

ezgif com-22e9fb136d

kunall17:
Btw awesome work on finding the bug in 921fb86 👍

Sam1301:
hmm.. its working an android 6.0.1 with chat.zulip.org. Can you please tell the android version on which you checked?

kunall17:
It shouldn't matter with the android version, anyways its 6.0.1!

Sam1301:
here's the [recording](https://drive.google.com/open?id=0B4jK5QX65b0ZanYxVXIyM2dYWDQ), working for me on android 6.0.1. Maybe some particular steps are involved to replicate this bug?

kunall17:
Hmmm, checked again works now but I get many logs of already have still! 
Is that normal?

kunall17:
@timabbott LGTM! :)

timabbott:
Cool, merged, thanks @Sam1301 and @kunall17!  I did fix one typo in the first line of the last commit message (`poistion` to `position`).

PR #283, opened 2017-01-07
Fix null streams

Sam1301:
EDIT:
Currently stream objects retrieved from messages are null throughout the project. I have corrected this by querying a stream object when we fetch new messages.

- listen for add, remove and update operations for subscription type events
- listen for muted topics events.
- in user configuration response, also get unsubscribed streams to ensure streams are not null in the message list.

### DIFF
--- a/app/src/main/java/com/zulip/android/models/Message.java
+++ b/app/src/main/java/com/zulip/android/models/Message.java
@@ -235,6 +235,11 @@ public class Message {
                     for (Message m : messages) {
                         Person person = Person.getOrUpdate(app, m.getSenderEmail(), m.getSenderFullName(), m.getAvatarUrl());
                         m.setSender(person);
+                        Stream stream = null;
+                        if (m.getType() == MessageType.STREAM_MESSAGE) {
+                            stream = Stream.getByName(app, m.getRecipients());
+                        }
+                        m.setStream(stream);
                         messageDao.createOrUpdate(m);
                     }
                     return null;
--- a/app/src/main/java/com/zulip/android/models/Stream.java
+++ b/app/src/main/java/com/zulip/android/models/Stream.java
@@ -39,7 +39,7 @@ public class Stream {
     private String description;
 
     @SerializedName("subscribers")
-    private List<Integer> subscribers;
+    private List<String> subscribers;
 
     @SerializedName("pin_to_top")
     private boolean pinToTop;
@@ -211,4 +211,12 @@ public class Stream {
         return null;
     }
 
+    public void setFetchColor(String fetchedColor) {
+        this.fetchedColor = fetchedColor;
+        getParsedColor();
+    }
+
+    public void setInHomeView(boolean inHomeView) {
+        this.inHomeView = inHomeView;
+    }
 }
--- a/app/src/main/java/com/zulip/android/activities/ExpandableStreamDrawerAdapter.java
+++ b/app/src/main/java/com/zulip/android/activities/ExpandableStreamDrawerAdapter.java
@@ -7,6 +7,7 @@ import android.widget.SimpleCursorTreeAdapter;
 
 import com.zulip.android.ZulipApp;
 import com.zulip.android.models.Message;
+import com.zulip.android.models.Stream;
 import com.zulip.android.util.ZLog;
 
 import java.sql.SQLException;
@@ -35,7 +36,7 @@ public class ExpandableStreamDrawerAdapter extends SimpleCursorTreeAdapter {
         try {
             results = ZulipApp.get().getDao(Message.class).queryRaw("SELECT DISTINCT subject, count(case when messages.id > " + pointer + " or messages." + Message.MESSAGE_READ_FIELD + " = 0 then 1 end) as unreadcount FROM messages " +
                     "JOIN streams ON streams.id=messages.stream " +
-                    "WHERE streams.id=" + groupCursor.getInt(0) + " group by subject").getResults();
+                    "WHERE streams.id=" + groupCursor.getInt(0) + " and streams." + Stream.SUBSCRIBED_FIELD + " = " + "1 group by subject").getResults();
         } catch (SQLException e) {
             ZLog.logException(e);
         }
--- a/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
+++ b/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
@@ -443,8 +443,8 @@ public class RecyclerMessageAdapter extends RecyclerView.Adapter<RecyclerView.Vi
                 } catch (SQLException e) {
                     ZLog.logException(e);
                 }
+                zulipApp.markMessageAsRead(message);
             }
-            zulipApp.markMessageAsRead(message);
         } catch (NullPointerException e) {
             Log.w("scrolling", "Could not find a location to scroll to!");
         }
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -1,13 +1,6 @@
 package com.zulip.android.activities;
 
 import android.Manifest;
-import java.sql.SQLException;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.concurrent.Callable;
-import java.util.ArrayList;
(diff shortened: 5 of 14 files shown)


smarx:

kunall17:
Awesome work @sam1301 but can you please put the second commit into a new PR!

Sam1301:
sure :+1:

Sam1301:
As stream objects will be added to each message, we would need to reset the existing database which has null values for streams in each message. Hence, I have suggested to bump up the database version which would reset the database and new messages will be fetched.

kunall17:
Nice work! 
Works as excepted. 

@timabbott LGTM, just FYI this upgrades the database version hence all the existing data are dropped and reloaded we have to take this big step because the streams saved in the messages table are null therefore which is causing problems in other features! 
There is no noticeable UI change as sam1301 stated no re-login required! 
Here is the [Databases.zip](https://github.com/zulip/zulip-android/files/694391/Databases.zip) (contains the old and the newly updated database after running the project through this branch) if any one wants to check/debug!

app/src/main/java/com/zulip/android/models/Message.java
```diff
@@ -419,6 +419,8 @@ public Void call() throws Exception {
                     for (Message m : messages) {
                         Person person = Person.getOrUpdate(app, m.getSenderEmail(), m.getSenderFullName(), m.getAvatarUrl());
                         m.setSender(person);
+                        Stream stream = Stream.getByName(app, m.getRecipients());
+                        m.setStream(stream);
```
timabbott: How does this work for messages sent to an individual rather than a stream?

Sam1301: Apologies, I totally missed that, after giving it some thought I think setting streams to null for private messages can be a potential solution.

timabbott:
OK; may be worth some careful testing; @kunall17 @saketkumar95 does one of you have time?

app/src/main/java/com/zulip/android/models/Message.java
```diff
@@ -484,7 +487,7 @@ public String getDisplayRecipient(ZulipApp app) {
             ArrayList<String> names = new ArrayList<>();
 
             for (Person person : people) {
-                if (person.id != app.getYou().id || people.length == 1) {
+                if (!person.getEmail().equals(app.getYou().getEmail()) || people.length == 1) {
```
Sam1301: I think the person objects all have id 0 because the user configuration response we receive is not deserialized into a person object with an id as it is supposed to be auto-generated.

kunall17: No, the person ID is auto generated so it shouldn't be the same and if you check the dumped databases i posted the person have unique ID's

Sam1301: thanks, made the changes :+1:

kunall17:
@vishwesh3 @saketkumar95 It would be great if you guys could help testing this big PR, this fixes the unread count + messages.streams columns?

yadav-rahul:
@kunall17 Can I test this PR?

kunall17:
@yadav-rahul Sure go ahead.. 👍

vishwesh3:
sure.

kunall17:
@Sam1301 The new bubble looks good for small numbers, but for double digits the round shape looks small! 
Also how about displaying the unread count in the toolbar like Zulip (12) ?

Sam1301:
@kunall17 Thanks, I will make the changes :+1:

Sam1301:
Update
1. resize bubbles based on counts
https://drive.google.com/open?id=0B4jK5QX65b0ZX2k0S1REc0lfbW8
2. show counts in toolbar
https://drive.google.com/open?id=0B4jK5QX65b0ZcWx4a0xNTTAwQlU

kunall17:
@sam1301 I think we should this PR, and fix unread counts in another one because we are facing issue if the pointer is very low from the latest message ID then the android app will not fetch those messages and hence we have lesser unread counts, which is not the right thing to display! 
So we can merge the fixing part? 
The expandable adapter will be up and working again!

Sam1301:
Since we only fetch 100 messages below the latest pointer, we won't get the correct number of unread counts for all the messages past that. Can we change this and fetch all the messages after the current pointer?

timabbott:
@Sam1301 we probably want to limit it in some way.  The backend is happy fetching up to ~1k messages, but if a user is somehow 25K messages behind, the request will fail.  

What we do on web is run a loop fetching in batches of 400 until we get to current.  We also offer the user the opportunity to "declare bankruptcy" if they have more than ~500 unread messages.

I think for practical purposes, once the user has over 1000 unread messages beyond the pointer, that's more than they're realistically going to scroll through anyway, so we could probably just stop at some point as a solution for now (there's not a cheap way to query the backend for grand total unread messages by recipient).

kunall17:
@Sam1301 Awesome Work! 👍 

@timabbott Now this PR has no unread counts commit's so we can merge this one! :)

app/src/main/java/com/zulip/android/models/Stream.java
```diff
@@ -120,9 +120,8 @@ public static Stream getByName(ZulipApp app, String name) {
             if (stream == null) {
                 Log.w("Stream.getByName",
                         "We received a stream message for a stream we don't have data for. Fake it until you make it.");
-                stream = new Stream(name);
-                app.getDao(Stream.class).createIfNotExists(stream);
-            }
+                // catch the new stream object with auto-generated id
+                stream = app.getDao(Stream.class).createIfNotExists(new Stream(name));            }
```
timabbott: Why do you need to generate IDs for Stream objects?  Streams have an ID on the backend...

I guess maybe fixing that would be harder to fix...

Sam1301: I think when stream is null, and a new stream object is created... using Stream(name) constructor, it doesn't get an id assigned to it and streams with id 0 are stored in the database with `createIfNotExists`, hence I tried to generate the ids. Am I mistaken?

timabbott: I should explain more.  I think your code is probably correct for how the Android app works today.  But what I think is going on here that is probably not ideal is that Zulip servers give streams an ID, and the Android app gives streams an ID (here), and those IDs have nothing to do with each other.  :(.

Sam1301: I know the stream objects have id given from server which are deserialized into id field in stream class here. I guess my concern is if the stream is null after querying for it in the database, we would be making a new stream object for it with id zero and store it in the database. Eventually, other stream objects generated this way would also have id 0 as well. Storing these in the database might cause a unique constraint exception. Am I mistake? I just want to clarify this doubt :)

timabbott: Well, storing a made-up stream ID for it might also cause a unique constraint exception if/when the stream that actually has that ID on the backend shows up, right?  More rare, but still a problem.  Hrm.

It seems like there are 2 possible correct solutions here:
* The same approach the web client uses, namely that we fetch from the server the complete list of streams we're subscribed to on startup, and maintain that list over time using the events system, so that this condition never happens.
* We modified the way messages are sent down to clients to optionally include the Stream ID in the display_recipient field (or something), so that the correct stream ID would be available here.

What do you think we should do?

Sam1301: I agree with the approach followed in the web client. That seems less susceptible to future bugs.

app/src/main/java/com/zulip/android/models/Stream.java
```diff
@@ -39,7 +39,7 @@
     private String description;
 
     @SerializedName("subscribers")
-    private List<Integer> subscribers;
+    private List<String> subscribers;
```
kunall17: These subscribers is the list of email right?

Sam1301: yes

kunall17:
Looks good to me now
LGTM @timabbott !

timabbott:
Merged, thanks @Sam1301!

app/src/main/java/com/zulip/android/models/Stream.java
Sam1301: Thanks for giving me such a nice explanation, I learned a lot from this pr :)

Issue #310, opened 2017-01-18
Narrow to a particular date doesn't show the messages posted on that day exclusively.


Sam1301:
When I narrow to a particular date using enter date, sometimes I see messages from previous day as well.

![screenshot_20170119-021640](https://cloud.githubusercontent.com/assets/14239866/22083967/314f7cfc-ddf4-11e6-9a80-e35e3357ea43.png)

yadav-rahul:
Yes @Sam1301. I agree with you. Actually there are many bugs in the PR made by @vishwesh3.

@kunall17 I think we should revert `Enter Date`  PR till all issues related to it are resolved?

vishwesh3:
Thanks @sam1301 for reporting. I will look into this and update soon.

vishwesh3:
I have solved this issue using #311. Can you please try this solution on your device and let me know if this issue still persists?

Sam1301:
tried the changes :+1:

kunall17:
@yadav-rahul what other issues are you facing?

yadav-rahul:
@kunall17 As @anuradhagupta345 mentioned in #303, `Enter Date` feature was not working in my app too.
Also when I go to particular date and then press `One Day Before` then also it was showing no messages.

kunall17:
@vishwesh3 
>Also when I go to particular date and then press One Day Before then also it was showing no messages.

vishwesh3:
Working on this and update soon.

vishwesh3:
I have solved and updated PR https://github.com/zulip/zulip-android/pull/311. Now date filtration from local DB is successfully working.Let me know if this still persists. 
Once https://github.com/zulip/zulip/issues/3406 is solved we will able to fetch messages of a specific date and will no longer be dependent on only local DB.

PR #311, opened 2017-01-19
Fix:310 Show messages of that date exclusively

vishwesh3:
PR addressing #310

### DIFF
--- a/app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
+++ b/app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
@@ -3,23 +3,26 @@ package com.zulip.android.filters;
 import android.os.Parcel;
 import android.os.Parcelable;
 
-import com.j256.ormlite.stmt.SelectArg;
 import com.j256.ormlite.stmt.Where;
+import com.zulip.android.R;
+import com.zulip.android.ZulipApp;
 import com.zulip.android.models.Message;
 import com.zulip.android.models.Stream;
+import com.zulip.android.util.Constants;
 
 import org.json.JSONException;
 
 import java.sql.SQLException;
 import java.text.SimpleDateFormat;
+import java.util.Calendar;
 import java.util.Date;
+import java.util.Locale;
 
 public class NarrowFilterByDate implements NarrowFilter {
 
     public static final Parcelable.Creator<NarrowFilterByDate> CREATOR = new Parcelable.Creator<NarrowFilterByDate>() {
         public NarrowFilterByDate createFromParcel(Parcel in) {
-
-            return new NarrowFilterByDate();
+            return new NarrowFilterByDate(new Date(in.readLong()));
         }
 
         public NarrowFilterByDate[] newArray(int size) {
@@ -27,6 +30,8 @@ public class NarrowFilterByDate implements NarrowFilter {
         }
     };
     private Date date = new Date();
+    private static Calendar calendar = Calendar.getInstance();
+    private static Calendar calendar2 = Calendar.getInstance();
 
     public NarrowFilterByDate() {
     }
@@ -35,44 +40,45 @@ public class NarrowFilterByDate implements NarrowFilter {
         this.date = date;
     }
 
-    /**
-     * Checks two dates are of same day or not
-     *
-     * @param date1 long date1 to be compared with date2
-     * @param date2 long date2 to be compared with date1
-     * @return boolean
-     */
-    private static boolean isSameDay(long date1, long date2) {
-        return date1 / 86400000 == date2 / 86400000;
-    }
-
     public int describeContents() {
         return 0;
     }
 
     public void writeToParcel(Parcel dest, int flags) {
+        dest.writeLong(date.getTime());
     }
 
     @Override
     public Where<Message, Object> modWhere(Where<Message, Object> where)
             throws SQLException {
-        where.like(Message.TIMESTAMP_FIELD, new SelectArg(date));
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(date);
+        calendar.set(Calendar.HOUR_OF_DAY, 0);
+        calendar.set(Calendar.MINUTE, 0);
+        calendar.set(Calendar.SECOND, 0);
+        calendar.set(Calendar.MILLISECOND, 0);
+
+        Date fromDate = calendar.getTime();
+        calendar.add(Calendar.DAY_OF_MONTH, 1);
+        Date toDate = calendar.getTime();
+
+        where.between(Message.TIMESTAMP_FIELD, fromDate, toDate);
         return where;
     }
 
     @Override
     public boolean matches(Message msg) {
-        return isSameDay(date.getTime(), msg.getTimestamp().getTime());
+        return isSameDay(date, new Date(msg.getTimestamp().getTime()));
     }
 
     @Override
     public String getTitle() {
-        return isSameDay(date.getTime(), new Date().getTime()) ? "Today Messages" : "Messages";
+        return isSameDay(date, new Date()) ? ZulipApp.get().getString(R.string.today_messages) : ZulipApp.get().getString(R.string.messages);
     }
 
     @Override
     public String getSubtitle() {
-        SimpleDateFormat dateFormat = new SimpleDateFormat("dd/MM/yyyy");
+        SimpleDateFormat dateFormat = new SimpleDateFormat(Constants.DATE_FORMAT, Locale.getDefault());
         return dateFormat.format(date);
     }
 
@@ -94,5 +100,19 @@ public class NarrowFilterByDate implements NarrowFilter {
     @Override
     public String toString() {
         return "{}";
+    }
+
+    /**
+     * Checks two dates are of same day or not
+     *
+     * @param date1 Date date1 to be compared with date2
+     * @param date2 Date date2 to be compared with date1
+     * @return boolean
+     */
+    private static boolean isSameDay(Date date1, Date date2) {
+        calendar.setTime(date1);
+        calendar2.setTime(date2);
+        return calendar.get(Calendar.YEAR) == calendar2.get(Calendar.YEAR) &&
+                calendar.get(Calendar.DAY_OF_YEAR) == calendar2.get(Calendar.DAY_OF_YEAR);
     }
 }
--- a/app/src/main/java/com/zulip/android/models/Message.java
+++ b/app/src/main/java/com/zulip/android/models/Message.java
@@ -20,6 +20,7 @@ import com.j256.ormlite.stmt.DeleteBuilder;
 import com.j256.ormlite.table.DatabaseTable;
 import com.zulip.android.R;
 import com.zulip.android.ZulipApp;
+import com.zulip.android.util.Constants;
 import com.zulip.android.util.CustomHtmlToSpannedConverter;
 import com.zulip.android.util.UrlHelper;
 import com.zulip.android.util.ZLog;
@@ -196,7 +197,7 @@ public class Message {
             this.setSubject(null);
         }
 
-        this.setTimestamp(new Date(message.getLong("timestamp") * 1000));
+        this.setTimestamp(new Date(message.getLong("timestamp") * Constants.MILLISECONDS_IN_A_MINUTE));
         this.setMessageRead(false);
     }
 
--- a/app/src/main/java/com/zulip/android/util/Constants.java
+++ b/app/src/main/java/com/zulip/android/util/Constants.java
@@ -1,0 +1,10 @@
+package com.zulip.android.util;
+
+/**
+ * List of all Constants used in the projects
+ */
+
+public class Constants {
+    public static int MILLISECONDS_IN_A_MINUTE = 1000;
+    public static String DATE_FORMAT = "dd/MM/yyyy";
+}
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -125,6 +125,8 @@
     <string name="SHOW">SHOW</string>
     <string name="notify_stream" formatted="false">" in %1$s/%2$s"</string>
     <string name="notify_private" formatted="false">from %1$s</string>
+    <string name="today_messages">Today Messages</string>
+    <string name="messages">Messages</string>
     <plurals name="new_message_mul_sender">
         <item quantity="one">%d new message</item>
         <item quantity="other">%d new messages</item>


smarx:

Sam1301:
Nice @vishwesh3 , works well :+1:

Sam1301:
just noticed the commit message, it doesn't include the hash `#` before issue number :)

kunall17:
- Include the # in the commit message
- Code not formatted! 

Otherwise LGTM!

vishwesh3:
Done

vishwesh3:
Now messages are filtered from local DB are shown. Filtering will be enhanced once https://github.com/zulip/zulip/issues/3406 is solved. 

https://drive.google.com/open?id=0BwZ3hknwHurYUDJObm0ybDZmUU0

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -56,23 +47,30 @@ public void writeToParcel(Parcel dest, int flags) {
     @Override
     public Where<Message, Object> modWhere(Where<Message, Object> where)
             throws SQLException {
-        where.like(Message.TIMESTAMP_FIELD, new SelectArg(date));
+        DateUtils.truncate(new Date(), java.util.Calendar.DAY_OF_MONTH);
```
kunall17: `DateUtils.truncate` returns an rounded date you are not using this why is this here?

https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/time/DateUtils.html#truncate(java.util.Calendar, int)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -56,23 +47,30 @@ public void writeToParcel(Parcel dest, int flags) {
...
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(date);
+        calendar.set(Calendar.HOUR_OF_DAY,0);
+        calendar.set(Calendar.MINUTE,0);
+        calendar.set(Calendar.SECOND,0);
+        calendar.set(Calendar.MILLISECOND, 0);
+        where.between(Message.TIMESTAMP_FIELD,calendar.getTime().getTime(),calendar.getTime().getTime()+86400000);
```
kunall17: Declare this 86400000 into a constant, this would help to understand this easier!

kunall17: Also format your code!

app/src/main/java/com/zulip/android/models/Message.java
```diff
@@ -104,7 +105,7 @@
     private String subject;
     @SerializedName("timestamp")
     @DatabaseField(columnName = TIMESTAMP_FIELD)
-    private Date timestamp;
+    private long timestamp;
```
kunall17: Changing the column fields isn't the best option, will it still work even for the old message table?

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
vishwesh3: Removed `DateUtils.truncate`. Forgot to remove it before

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
vishwesh3: Done

app/src/main/java/com/zulip/android/models/Message.java
vishwesh3: As column field is changed so DB version is increased and onUpdate method is called. Thus it working for old messages too. :)

Sam1301: @vishwesh3 just trying to understand the code, can you please tell why changing the type of timestamp in message model is needed?

vishwesh3: Let's take a example (assume that back dated messages are stored in local DB (will be fine once https://github.com/zulip/zulip/issues/3406 is done) and today's date is  25/01/2017)

Now when we narrow to date let's say 20/01/2017. 
**Before this PR**
Database query is done with `where.like(Message.TIMESTAMP_FIELD, new SelectArg(date));` (https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java#L59) rather than filtering messages of that specific date it just do nothing. When I debugged I found that list contains messages of all dates, but messages of other days ( except 20/01/2017) are not shown because it returns false on match (https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java#L65). All DB quires are done with limit of 100. 
**Case 1**
So if messages from 20 to 25 are less than 100 thus messages returned by DB query includes messages of 20 and only messages dated 20 are shown with the help of match method. 
**Case 2**
Messages from 20 to 25 are greater than 100 thus messages returned by DB query does not include messages of 20, even if they are stored in the local DB. Also no messages are displayed so scrolling is not possible and no further DB quires takes place.

**Problem** Where clause is not working as per expectation.

**In this PR**
Changed where clause (https://github.com/vishwesh3/zulip-android/blob/narrow/app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java#L57) 
Now DB query returns only those messages whose timestamp is in between narrowed date.

Sam1301: @vishwesh3 I really appreciate how well you explained your point :) Although I still feel that changing the type of timestamp for a message is unnecessary. Ormlite [where.between](http://ormlite.com/javadoc/ormlite-core/com/j256/ormlite/stmt/Where.html#between(java.lang.String,%20java.lang.Object,%20java.lang.Object)) can take any type of data (in this case Date()) to return a modulated where. So I tried these [changes](https://gist.github.com/Sam1301/90a7271024e2b560d46e7f885a332198) to NarrowFilterByDate.java and it worked for me. The code in my example could obviously be more concise but I guess you get the point :)

vishwesh3: Thanks @Sam1301 I will play around and will update soon.

vishwesh3: Done :)

vishwesh3:
Done :) 
Please have a final look.

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -56,23 +46,37 @@ public void writeToParcel(Parcel dest, int flags) {
...
+        calendar.set(Calendar.MILLISECOND, 0);
+
+        Calendar calendar1 = Calendar.getInstance();
+        calendar1.setTime(date);
+        calendar1.set(Calendar.HOUR_OF_DAY, 0);
+        calendar1.set(Calendar.MINUTE, 0);
+        calendar1.set(Calendar.SECOND, 0);
+        calendar1.set(Calendar.MILLISECOND, 0);
+        calendar1.add(Calendar.DAY_OF_MONTH, 1);
+        where.between(Message.TIMESTAMP_FIELD, calendar.getTime(), calendar1.getTime());
```
Sam1301: there is some code duplication here, can be reduced

vishwesh3: Done.

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -56,23 +47,33 @@ public void writeToParcel(Parcel dest, int flags) {
...
+        calendar.setTime(date);
+        calendar.set(Calendar.HOUR_OF_DAY, 0);
+        calendar.set(Calendar.MINUTE, 0);
+        calendar.set(Calendar.SECOND, 0);
+        calendar.set(Calendar.MILLISECOND, 0);
+
+        Calendar calendar1 = Calendar.getInstance();
+        calendar1.setTime(calendar.getTime());
+        calendar1.add(Calendar.DAY_OF_MONTH, 1);
+        where.between(Message.TIMESTAMP_FIELD, calendar.getTime(), calendar1.getTime());
```
Sam1301: ok so here we are creating 2 Calendar instances. I think we could use just one calendar instance
Example:
```
date1 = calendar.getTime();
calendar.add(DAY_OF_MONTH, 1);
date2 = calendar.getTime();
```

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -3,16 +3,18 @@
...
 
 import org.json.JSONException;
 
 import java.sql.SQLException;
 import java.text.SimpleDateFormat;
+import java.util.Calendar;
 import java.util.Date;
+import java.util.Locale;
 
 public class NarrowFilterByDate implements NarrowFilter {
```
Sam1301: Also I just noticed, that we are not using writeToParcel method in NarrowFilterByDate. When we need to pass objects across activities we implement the [Parcelable interface](https://developer.android.com/reference/android/os/Parcelable.html) for those object classes. We are saving NarrowFilter instances in a [bundle](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/MessageListFragment.java#L81) to save the state. See [NarrowFilterStream](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/filters/NarrowFilterStream.java) for example.

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
vishwesh3: Done :)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -35,44 +37,45 @@ public NarrowFilterByDate(Date date) {
...
-    private static boolean isSameDay(long date1, long date2) {
-        return date1 / 86400000 == date2 / 86400000;
-    }
-
     public int describeContents() {
         return 0;
     }
 
     public void writeToParcel(Parcel dest, int flags) {
+        dest.writeValue(date);
```
Sam1301: When you write data to parcel, it is analogous to serialization, so we also need to deserialize the result to get back the NarrowFilterByDate instance in [creator createfromparcel](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java#L19).

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -3,23 +3,25 @@
...
+import java.util.Locale;
 
 public class NarrowFilterByDate implements NarrowFilter {
 
     public static final Parcelable.Creator<NarrowFilterByDate> CREATOR = new Parcelable.Creator<NarrowFilterByDate>() {
         public NarrowFilterByDate createFromParcel(Parcel in) {
-
-            return new NarrowFilterByDate();
+            in.setDataPosition(0);
+            return new NarrowFilterByDate((Date) in.readValue(Date.class.getClassLoader()));
```
vishwesh3: Deserialized the result to get back the NarrowFilterByDate instance. :)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
vishwesh3: Done :)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
Sam1301: Awesome :) Only one comment though, we don't need to set data position to initial since its already set in the beginning when createFromParcel is called :)

Sam1301: btw just a suggestion, by using writeValue(Date object) we are using general serialization which is kinda slow pref wise against parcelable, and that is the whole point why in android parcelable is preferred over general serialization(https://developer.android.com/reference/android/os/Parcel.html#writeValue(java.lang.Object))

Sam1301:
@vishwesh3 also can you please squash the two commits as they both address the same problem? Maybe even add more description to commit message explaining "why" the change is needed :)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
vishwesh3: Used long and created Date with it at the time of deserialization. :)

Sam1301:
@vishwesh3 awesome, can you please fix some typos in the commit message :)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
Sam1301: cool :)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -35,44 +36,45 @@ public NarrowFilterByDate(Date date) {
...
     @Override
     public boolean matches(Message msg) {
-        return isSameDay(date.getTime(), msg.getTimestamp().getTime());
+        return isSameDay(date, new Date(msg.getTimestamp().getTime()));
     }
 
     @Override
     public String getTitle() {
-        return isSameDay(date.getTime(), new Date().getTime()) ? "Today Messages" : "Messages";
+        return isSameDay(date, new Date()) ? "Today Messages" : "Messages";
```
kunall17: User displayed strings in strings.xml!

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
```diff
@@ -95,4 +97,20 @@ public String getJsonFilter() throws JSONException {
...
+     * Checks two dates are of same day or not
+     *
+     * @param date1 Date date1 to be compared with date2
+     * @param date2 Date date2 to be compared with date1
+     * @return boolean
+     */
+    private static boolean isSameDay(Date date1, Date date2) {
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(date1);
+        Calendar calendar2 = Calendar.getInstance();
```
kunall17: This isSameDay will be called very frequently via https://github.com/zulip/zulip-android/blob/a4374c12240d3714710f2dac4376dbecc692a49c/app/src/main/java/com/zulip/android/activities/MessageListFragment.java#L335 
So i'm kinda confused is this getInstance returning a new object? 
If so it would be better to cache these two objects!

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
vishwesh3: Done :)

app/src/main/java/com/zulip/android/filters/NarrowFilterByDate.java
vishwesh3: Done :)

kunall17:
@timabbott LGTM!

timabbott:
@vishwesh3 I cleaned up the typos in your commit message and merged this as 41ceacd2522a4b727f7455b74afb4e55d2f401f3.

Thanks for fixing this!!

Issue #313, opened 2017-01-20
Adding templates for issues and Pull Requests 


vishwesh3:
https://github.com/blog/2111-issue-and-pull-request-templates

This will make sure that important information is not missed in issue.
**Points in issue template**
- Actual Behaviour
- Expected Behaviour
- Steps to reproduce it
- LogCat for the issue
- Screenshots of the issue
- Would you like to work on the issue?

**Check box in Pull Request template**

- [ ]  Code is formatted Mac: `Option + Command + L` Windows: `Ctrl + Alt + L` Linux: Go to `File->Settings->(under IDE settings)->Keymap` and change shortcut as `Ctrl + Alt + L` locks linux.
- [ ]  Make sure BUILD is SUCCESSFUL after running `./gradlew lintDebug` in terminal

vishwesh3:
@timabbott @kunall17 @Sam1301 your views on this?

kunall17:
Hmm, interesting 
Also a check if the PR is related to UI then it works, shows well for the black theme that part misses frequently!

vishwesh3:
Working on this and create PR soon :)

PR #314, opened 2017-01-20
Added Templates for Issue and Pull Requests

vishwesh3:
Issue #313

Let me know if changes are required.


smarx:

timabbott:
Both of these, but especially the issue template, feels really long; I worry about them intimidating potential bug reporters with what feels like a bunch of paperwork.  I think it's probably worth moving a lot of the content (especially e.g. the instructions for how to format code) into another document in the repo that we can link to, and trying to really simplify these to the most important information that we need.

kunall17:
@timabbott Yeah I agree for the issue part they are really long but we should include the checkList for the PR section! 

@vishwesh3 Also include for the usage of` Zlog.logException(e)` instead of `e.printStackTrace`

vishwesh3:
Created new files for guide([Issue Guide](https://github.com/vishwesh3/zulip-android/blob/adding-templates/.github/ISSUES_GUIDE.md) and [Pull Request Guide](https://github.com/vishwesh3/zulip-android/blob/adding-templates/.github/PULL_REQUEST_GUIDE.md)) and linked them with template.

timabbott:
@kunall17 perhaps we should make a linter check for use of `e.printStackTrace` and tell users to use `Zlog.logException` instead?

kunall17:
@timabbott yeah that would be better, will have to write a custom lint rules!

vishwesh3:
Yes, it would be better. I will update PR soon. :)

kunall17:
@Sam1301 can you give a review on 4dc35e5 ?

vishwesh3:
Done, Now if `e.printStackTrace` is used then it is shown in lint report like this and build is failed.
screenshot from 2017-01-24 19-43-32

kunall17:
Looks good, it would be better to include some info about why are we forcing to use Zlog!

customLintAAR/src/test/java/com/zulip/android/customlintarr/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
+package com.zulip.android.customlintarr;
```
Sam1301: unless this file is used to test the custom lint rule, I think it can be deleted.

customLintAAR/src/androidTest/java/com/zulip/android/customlintarr/ApplicationTest.java
```diff
@@ -0,0 +1,13 @@
+package com.zulip.android.customlintarr;
```
Sam1301: same comment (I think it can be deleted)

Sam1301:
@vishwesh3 Awesome work! :+1:  posted a comment on custom lint rule commit.

vishwesh3:
@kunall17 Added more info. Let me know if some more any specific info you want to add. :)
@Sam1301 Done. I have removed those testing files, will add once we start adding test for lint rules. :)

kunall17:
LGTM! :)
@Sam1301 Can you give this a final review?

vishwesh3:
**Update**
Removed checkBox for `ZLog` ( Used `Zlog.logException(e)` instead of `e.printStackTrace()`) as new rule is added in `./gradlew lintDebug`

Sam1301:
@kunall17 reviewed, @vishwesh3 great...the custom lint rule commit looks good to me :+1:

kunall17:
@timabbott What about this one, It's good we'll have a template for the PR's and issues where it will not be mandatory to follow it, but if they do its good! 
LGTM if you find this ok!

timabbott:
I edited the copy somewhat (preserving the checklist) and merged this.  Huge thanks for working on this @vishwesh3!  I think we'll probably want to do some further tweaks as we see this in action; it's possible that e.g. the issue template is too heavy.

One thing to probably fix right away is that the link to the issue guide in the issue template is kinda hard to use.

Thanks @vishwesh3!

vishwesh3:
Yes, we can make further changes, in order to improve. :+1:

Issue #309, opened 2017-01-18
Small issues in UI 


vishwesh3:
1. On narrowing to our self title is not set. 
![videotogif_2017 01 18_09 11 04](https://cloud.githubusercontent.com/assets/18511177/22051649/bdd613c8-dd6a-11e6-9400-52a066e9fb00.gif)

2. On narrowing to PM it will be better to automatically set Person field.
And on narrowing to stream set stream field in message dialog, on narrowing to specific topic of stream set stream name and topic. 
![videotogif_2017 01 18_09 14 49](https://cloud.githubusercontent.com/assets/18511177/22051695/f72a0576-dd6a-11e6-980c-156da620e1e6.gif)
 

3. When drawers are open and if user presses back button it should first close drawer. 
![videotogif_2017 01 18_09 18 44](https://cloud.githubusercontent.com/assets/18511177/22051767/9ae9994c-dd6b-11e6-9fd1-f38305b2c47c.gif)

4. In home if user presses back button twice it finishes app but toast is still showing, which doesn't look good. (above GIF)

5. If messages list is less and doesn't occupy full screen then on scroll toolbar toggle, which doesn't look good. It will be better to restrict toolbar to toggle if message list is less (doesn't occupy full screen)  
![videotogif_2017 01 18_09 20 25](https://cloud.githubusercontent.com/assets/18511177/22051882/55079b76-dd6c-11e6-8e24-f880f9664de5.gif)
 
6. On logging to localhost server after login, login activity is not finished.
![videotogif_2017 01 18_09 22 32](https://cloud.githubusercontent.com/assets/18511177/22051979/14228b7e-dd6d-11e6-88fa-737f2c93f1a9.gif)

7. If the entered server url is wrong cursor of the edittext is set to start, it will be better to set it at last so that user can easily correct url.(pressing back button)

8. When new message is arrived snackbar is show whose start animation is from bottom, which should be from top.(as we are showing snackbar at the top) 
![videotogif_2017 01 18_09 07 44](https://cloud.githubusercontent.com/assets/18511177/22051634/a88bb7b6-dd6a-11e6-8d70-04dae38b78cf.gif)

I think this are bugs, let me know if they are intentionally done like that.
All this are small issue and will don't take much time and should be included in our next upcoming release.  
@timabbott @kunall17 @Sam1301 your views on this

Sam1301:
[...]

PR #317, opened 2017-01-21
Fixed minor UI issues #309

vishwesh3:
PR addressing #309

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -396,7 +396,9 @@ public class ZulipActivity extends BaseActivity implements
                 if (id == allPeopleId) {
                     doNarrow(new NarrowFilterAllPMs(app.getYou()));
                 } else {
-                    narrow_pm_with(Person.getById(app, (int) id));
+                    Person person = Person.getById(app, (int) id);
+                    narrowPMWith(person);
+                    switchToPrivate();
                 }
             }
         });
@@ -569,6 +571,7 @@ public class ZulipActivity extends BaseActivity implements
                 if (narrowedList == null) {
                     calendar = Calendar.getInstance();
                     menu.getItem(2).getSubMenu().getItem(0).setTitle(R.string.menu_today);
+                    switchToStream();
                     checkForChatBoxFocusRequest();
                 } else if (narrowedList.filter instanceof NarrowFilterByDate) {
                     menu.getItem(2).getSubMenu().getItem(0).setTitle(R.string.menu_one_day_before);
@@ -578,7 +581,7 @@ public class ZulipActivity extends BaseActivity implements
     }
 
     private void checkForChatBoxFocusRequest() {
-        if (TextUtils.isEmpty(streamActv.getText().toString())) {
+        if (TextUtils.isEmpty(streamActv.getText().toString()) && isCurrentModeStream()) {
             streamActv.requestFocus();
         } else if (TextUtils.isEmpty(topicActv.getText().toString())) {
             topicActv.requestFocus();
@@ -1083,6 +1086,7 @@ public class ZulipActivity extends BaseActivity implements
             showView(chatBox);
         } else {
             hideView(chatBox);
+            hideSoftKeyBoard();
         }
     }
 
@@ -1203,6 +1207,7 @@ public class ZulipActivity extends BaseActivity implements
                 }
                 expandableListView.expandGroup(position);
                 previousClick = position;
+                onNarrowFillSendBoxStream(streamName, "", false);
                 return true;
             }
         });
@@ -1743,8 +1748,9 @@ public class ZulipActivity extends BaseActivity implements
         doNarrow(new NarrowFilterStream(stream, null));
     }
 
-    private void narrow_pm_with(final Person person) {
         doNarrow(new NarrowFilterPM(Arrays.asList(app.getYou(), person)));
+    private void narrowPMWith(final Person person) {
+        onNarrowFillSendBoxPrivate(new Person[]{person},false);
     }
 
     @TargetApi(Build.VERSION_CODES.HONEYCOMB)
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -1748,8 +1748,12 @@ public class ZulipActivity extends BaseActivity implements
         doNarrow(new NarrowFilterStream(stream, null));
     }
 
-        doNarrow(new NarrowFilterPM(Arrays.asList(app.getYou(), person)));
     private void narrowPMWith(final Person person) {
+        List<Person> list = new ArrayList<>();
+        list.add(person);
+        if (!person.getEmail().equals(app.getYou().getEmail()))
+            list.add(app.getYou());
+        doNarrow(new NarrowFilterPM(list));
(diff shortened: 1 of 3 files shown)


smarx:

yadav-rahul:
@vishwesh3 Seems like merge conflicts have been raised. Can you resolve them?

kunall17:
Remove these spacing in the corners 

device-2017-01-24-104743

vishwesh3:
@kunall17 Done Removed spaces. Some screenshots
screenshot from 2017-01-25 15-17-11
screenshot from 2017-01-25 14-47-31
screenshot from 2017-01-25 14-46-37

vishwesh3:
@yadav-rahul Solved merge conflicts

app/src/main/java/com/zulip/android/filters/NarrowFilterPM.java
```diff
@@ -73,8 +73,12 @@ public boolean matches(Message msg) {
     public String getTitle() {
         ArrayList<String> names = new ArrayList<>();
         for (Person person : people) {
-            if (!person.equals(ZulipApp.get().getYou()) || people.size() == 1) { //If PM to self then show title as your name
```
Sam1301: hmm... I think the comment here indicates that if `people.size()` is one, it means the narrow is with the person itself. The only problem is when user narrows to itself, the people list has two person objects both pointing to user itself. This is because https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L1686 is missing a condition where if `person.equals(app.getYou())` then we might just want to pass an arraylist with one person object i.e app.getYou(). So, I think we could reduce all this code to just an if condition in narrow_with_pm().

Sam1301:
@vishwesh3 If it isn't much trouble can you split all these fixes in their own prs. That way it will be easier to make changes and merge these fixes separately.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -391,7 +391,10 @@ public void onItemClick(AdapterView<?> parent, View view,
                 if (id == allPeopleId) {
                     doNarrow(new NarrowFilterAllPMs(app.getYou()));
                 } else {
-                    narrow_pm_with(Person.getById(app, (int) id));
+                    Person person = Person.getById(app, (int) id);
+                    narrow_pm_with(person);
+                    switchToPrivate();
+                    setRecipientsInChatBox(person.getEmail());
```
Sam1301: Nice, the code could be reduced though. ZulipActivity implements [NarrowListener](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/filters/NarrowListener.java) which has a function onNarrowFillSendBoxPrivate to perform this job :)

Also on narrowing to stream, the chatbox is not populated when I click on views except [stream name textview](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L1152)

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -190,7 +190,13 @@ public Void call() throws Exception {
                             } else {
                                 personDao.update(person);
                             }
-                        } catch (Exception e) {
+                            if (person.getEmail().equals(app.getEmail())) {
```
vishwesh3: This is also required as when we refresh (from menu) name in `app.getYou()` becomes null. To update name this is required.

app/src/main/java/com/zulip/android/filters/NarrowFilterPM.java
vishwesh3: Yes done in this way. :)

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
vishwesh3: Done :) Now I have used NarrowListener to set text. Thanks @Sam1301

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
Sam1301: @vishwesh3 I debugged and found that we are calling [setEmail](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/ZulipApp.java#L410) in [onRefresh](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L2109). Can you please check again and see if you are still getting this error?

vishwesh3: Yes we are calling `setEmail` method, but when I debugged I found that `personCache != null` is false as we are passing `personCache` as null [(here)](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/models/Person.java#L178) then  as `person` is null [(here)](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/models/Person.java#L149), cant get person[(here)](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/models/Person.java#L150) from DB as DB was reset just before `setEmail` call [(here)](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L2117) 
. [At](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/models/Person.java#L156) this point person is null  now `new Person()` is created with `email, name, avatarURL` where `name` and `avatarURL` are passed as null [(here)](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/ZulipApp.java#L412). I removed this lines and used [person.equals](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/models/Person.java#L281) and found that title is again blank. Am I missing out something?

Sam1301: @vishwesh3 yes, I agree, the name and avatarUrl would be null immediately after reset. In that case I think it'd be reasonable to check for email equality `if (person.getEmail().equals(app.getYou.getEmail()))` in `narrow_pm_with()`. If we go with this, I think we won't need this change in asyncGetEvents.java as the Person list would always contain 1 object referring to the user itself in case of narrow to self.

vishwesh3: Ok, will use `(person.getEmail().equals(app.getYou.getEmail()))` in `narrow_pm_with()`. But I think for future aspect we should update `you` in the `app`. Shall I create different PR for this update part?

Sam1301: yeah that'd be reasonable :+1:

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1737,7 +1737,11 @@ protected void narrow(final Stream stream) {
     }
 
     private void narrow_pm_with(final Person person) {
```
kunall17: Mostly the code base is following camelCase so can you change this?

vishwesh3: Done

Sam1301:
@vishwesh3 great, the code looks good to me :+1:

timabbott:
Merged after tweaking the commit message a bit, thanks @vishwesh3!

Issue #329, opened 2017-01-26
Open keyboard when starting composing a reply


timabbott:
When you touch a message body to reply, we should probably open the keyboard in order to save 1 click in starting a reply (one of the most common things folks will be doing).

vishwesh3:
Will send a PR soon.

PR #333, opened 2017-01-26
Fix:#329 Pop keyboard on Fab click

vishwesh3:
PR addressing issue #329 
Also pop when message is clicked at home

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -1042,8 +1042,24 @@ public class ZulipActivity extends BaseActivity implements
                 displayChatBox(true);
                 displayFAB(false);
                 fabHidder.start();
+                showSoftKeyboard();
+                if (TextUtils.isEmpty(streamActv.getText().toString())) {
+                    streamActv.requestFocus();
+                } else if (TextUtils.isEmpty(topicActv.getText().toString())) {
+                    topicActv.requestFocus();
+                } else {
+                    messageEt.requestFocus();
+                }
             }
         });
+    }
+
+    /**
+     * Shows soft keyboard
+     */
+    private void showSoftKeyboard() {
+        InputMethodManager imm = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
+        imm.toggleSoftInput(InputMethodManager.SHOW_IMPLICIT,InputMethodManager.HIDE_IMPLICIT_ONLY);
     }
 
     private void displayChatBox(boolean show) {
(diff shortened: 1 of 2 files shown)


smarx:

Sam1301:
@vishwesh3 this looks great :+1: , the code looks fine to me.

vishwesh3:
**Update**
- Check which editText to give focus
- onPause hide keyboard (if user presses home button or switch to another app)

@Sam1301 Can you please review updated changes? :)

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1042,10 +1042,20 @@ public void onClick(View view) {
...
             }
         });
     }
 
+    /**
+     * Shows soft keyboard
+     */
+    private void showSoftKeyboard() {
+        InputMethodManager imm = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
+        imm.toggleSoftInput(InputMethodManager.SHOW_FORCED,InputMethodManager.HIDE_IMPLICIT_ONLY);
```
Sam1301: I think we won't need the second commit if we use [SHOW_IMPLICIT](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager.html#SHOW_IMPLICIT) instead of SHOW_FORCED.

vishwesh3: Done :)

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1042,10 +1042,26 @@ public void onClick(View view) {
                 displayChatBox(true);
                 displayFAB(false);
                 fabHidder.start();
+                showSoftKeyboard();
+                if (streamActv.getText().toString().isEmpty() || streamActv.getText().toString().equals("")) {
```
Sam1301: a better way to check if string is null or empty is [TextUtils.isEmpty()](https://developer.android.com/reference/android/text/TextUtils.html)

vishwesh3: Done :)

vishwesh3:
In https://github.com/zulip/zulip-android/pull/333/commits/8822c23ab9560150c22ad8cacc66dff84fe0a709 covered corner case :- If user is narrowed to any stream (but not to topic) then focus is on topic and on back pressed home fragment is popped thus focus should come to stream. 

@Sam1301 Can you please review 2nd one? :)

Sam1301:
@vishwesh3 great, the first commit message could be more descriptive as it also shows keyboard when message is clicked. Can you please fix some typos in the second commit message? Checkout this example for a well descriptive [commit message](https://github.com/zulip/zulip/commit/e7091018d5ff3a084cdd15321ca86b6270f71e0c). Also be sure to checkout the commit message [guide](https://zulip.readthedocs.io/en/latest/version-control.html#commit-messages) :)

vishwesh3:
Done :)

timabbott:
Thanks for fixing this @vishwesh3!!

@vishwesh3 you should include periods at the end of the sentences in your commit messages.  I merged this after a bit more commit message cleanup; see 116a21eeb47a9c76afa91afbf18ed620921d939c for what I did with your first commit, for example.  Do you see how that version communicates much more clearly to people reading the commit history why this change was made?

vishwesh3:
Yes, new commit message is much more clear and helps others to know what is in this commit. It really improved version communication. Thanks, I will keep that in mind while committing my next commit.

PR #346, opened 2017-01-29
Fixes #325: Closes Soft Keyboard if it opens on activity recreation

### LINKED ISSUE #325: Keyboard keeps popping up when Theme is changed after typing a message
The keyboard does not hide after there is a change in theme when the user is typing the message. Then this keeps up in a cycle as and when i keep changing the theme.

g_20170125_0222423

@kunall17 @timabbott . I'd like to work on this.

nitish1211:
Fixes #325 

Here I am closing the keyboard if it remains open at startup.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -56,6 +56,7 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewPropertyAnimator;
+import android.view.WindowManager;
 import android.view.animation.Interpolator;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.AdapterView;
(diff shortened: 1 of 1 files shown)


smarx:

@nitish1211, it looks like you've already signed the Dropbox CLA. Thanks!

kunall17:
Already there is a method https://github.com/nitish1211/zulip-android/blob/bcc4643fff3e8f80e38b5a07baa7b6d48c610657/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L615 use this one!

nitish1211:
It does not work because `view=null` as the focus is not on any view(Usually when we use that method the focus is on the editText which is using it hence view wont be null) .So this check https://github.com/nitish1211/zulip-android/blob/bcc4643fff3e8f80e38b5a07baa7b6d48c610657/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L618 fails

kunall17:
@nitish1211 ok makes sense! 

@timabbott LGTM!

timabbott:
Merged as d3868cb4efbbcbb46680144a20aca428fc8b2e7a after tweaking the commit message style, thanks @nitish1211!

PR #348, opened 2017-01-30
Update you in ZulipApp after refresh.

vishwesh3:
This PR update `you` in the `zulipApp` whose fields becomes null after refresh as discussed in [this](https://github.com/zulip/zulip-android/pull/317#discussion_r98337381).


smarx:

Sam1301:
With the merge of #362 , this case will be automatically handled since `app.you` in that case always has a valid id (server person id).

timabbott:
@Sam1301 does that mean we should close this?

Sam1301:
@timabbott yeah, this is fixed with the merge of #362 :+1:

Issue #344, opened 2017-01-29
Adding a message at the end of all message list for user awareness.


harshitagupta30:
Just a suggestion can we add a message like "No new messages" at the end of the message list when the user is trying to update the list to make user aware that list has been updated instead of the present scenario shown in gif.

![videotogif_2017 01 29_20 24 52](https://cloud.githubusercontent.com/assets/12701036/22405034/2a54e128-e662-11e6-93b1-a5723d4a75c7.gif)

@timabbott @kunall17 @Sam1301 Hi, please review!

yadav-rahul:
@harshitagupta30 I don't think that adding "No new messages" at the end of list(while updating) would be a good option because How can one predict that whether new messages are there or not while updating the list?

harshitagupta30:
Since we're fetching the messages from API and if all messages have been loaded and after that user tries to fetch by scrolling then a message should pop out that all messages has been fetched.

kunall17:
I thought of implementing a snack bar or toast 
while writing the recycler view but the problem is it can be spammy as well! And may not look good. 

What are the other options for notifying?

harshitagupta30:
Hi @kunall17! 
In my opinion, it doesn't look spammy. Assuming one has low internet connection and he is trying to load new messages which is taking time and if they are not loaded he'll think the internet is not working properly while the case is all the messages has been loaded. So if we can provide a message it will let user know that all messages have been fetched.  
In fact I tried to implement that and it looks good to me. I am attaching the gif as well. 

![videotogif_2017 01 30_18 46 42](https://cloud.githubusercontent.com/assets/12701036/22424610/410c0944-e71d-11e6-950f-fee78273c590.gif)

What's your opinion?

I would love to hear your thoughts on this! @timabbott @Sam1301

kunall17:
@harshitagupta30 Yeah i can connect very well, and understand the problem! 
I also asked what can be other options rather than snackbar, if no other good suggestions we'll have to go with snackbar!  

Can you upload the code (maybe a .patch file) or a commit, I wanted to test?

harshitagupta30:
I tested with toast and alert dialog box as well but this the best possible way and very clean, in my opinion. 
Yeah! I will upload the .patch file for the testing.

Sam1301:
@harshitagupta30 feel free to claim this task using zulipbot. That way it will be helpful for others to skip through issues which are not already worked on :)

Sam1301:
[...]

PR #358, opened 2017-02-01
Issue#344 Snackbar added for user awareness resolved

harshitagupta30:
Fixes issue #344 
@kunall17 Please review!


smarx:

@harshitagupta30, it looks like you've already signed the Dropbox CLA. Thanks!

kunall17:
Snackbar overlaps the FAB can you fix this?

snackbar fab

harshitagupta30:
didn't notice that! Yeah, I'll fix this.

harshitagupta30:
Hi @kunall17! the best solution to this problem I could find is to hide the fab whenever snackbar appears and as soon as snackbar disappears fab will appear which doesn't overlap at all and looks good!  

videotogif_2017 02 02_16 12 51

kunall17:
How about using the animations which are currently being used rather than this fading?

harshitagupta30:
I didn't get your point.

kunall17:
https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L1100

harshitagupta30:
@kunall17 I earlier tried to use these animations but they resulted into unusual behaviour of the fab! Sometimes it appears and sometimes it doesn't! It also happened that it was slightly visible at the bottom but not complete. That is why I opted out for adding the new behavior.

kunall17:
Weird this method is being used at many places for animation hiding and stuff
The point is two animations for hiding the FAB isn't looking good one fadeout and one sliding!

Any possibility for sliding this FAB out?

harshitagupta30:
@kunall17 According to the documentation everywhere whenever we have snackbar along with the fab in coordinator layout then when the snackbar appears the fab will itself move up and tried that approach but not got that result. Then use the animation which was resulting into the fab being half hidden and half visible and then I came out with this i.e. hiding the whenever the snackbar pops out.

kunall17:
@harshitagupta30 yeah it would have if we were using the default **behavior** of the fab and the coordinator layouts! 
Here we use custom one's hence this does not works!

harshitagupta30:
@kunall17 Okay! Do I need to make any changes in this? I don't think its looking bad! :)

harshitagupta30:
@kunall17 Can you please review this and finalise the changes in this PR so that this issue can be resolved?
Thanks!

kunall17:
Still overlapping 
device-2017-02-07-030901

app/src/main/java/com/zulip/android/util/ShrinkBehaviorOfFab.java
```diff
@@ -0,0 +1,72 @@
...
+import android.support.design.widget.CoordinatorLayout;
+import android.support.design.widget.FloatingActionButton;
+import android.support.design.widget.Snackbar;
+import android.support.v4.view.ViewCompat;
+import android.util.AttributeSet;
+import android.view.View;
+
+import java.util.List;
+
+/**
```
kunall17: Remove this author details, the git history has the author details!

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -170,6 +174,14 @@ public void onScrollStateChanged(RecyclerView recyclerView, int newState) {
     }
 
     @Override
+    public void onViewCreated(View view, @Nullable Bundle savedInstanceState) {
+        super.onViewCreated(view, savedInstanceState);
+        CoordinatorLayout coordinatorLayout = (CoordinatorLayout) getActivity().findViewById(R.id.coordinatorLayout);
+        snackbar = Snackbar.make(coordinatorLayout,R.string.no_new_messages,Snackbar.LENGTH_SHORT)
```
kunall17: Format the whole code (ctrl + alt + L)

kunall17:
This is still overlapping because i think your behaviour is not being used now, because behaviour is set here by java https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/ZulipActivity.java#L1674

harshitagupta30:
Hi @kunall17! I've made the changes as told by you and yes that behavior was not working! I've added the required code and I guess its working fine now! Can you please review it? :)

harshitagupta30:
@kunall17 Can you please review it again and let me know if any changes required? :+1:

brainwane:
@harshitagupta30 Thanks for your pull request, and thanks for writing a descriptive commit message. [Here's a guide to help you write commit messages](https://zulip.readthedocs.io/en/latest/version-control.html#commit-messages) that are more consistent with our style, which makes it easier for us in the future -- when someone's trying to solve a new problem, if commit messages are all in the same format, it's easier to skim them in logs and quickly understand what's happening.

[Here are some tips on tidying your commit history, including changing your commit messages.](https://zulip.readthedocs.io/en/latest/git-guide.html#examine-and-tidy-your-commit-history) And when you update your code or your commit message, you can still of course [update your pull request](https://zulip.readthedocs.io/en/latest/git-guide.html#update-a-pull-request) and the changes will appear here. :)

Thanks!

harshitagupta30:
@kunall17 Can you please review it? :+1:

harshitagupta30:
@kunall17 I am referring the same article! I found it much earlier and was tried to apply the same thing here :D But alas it's not working here! 
Now, what should I do? any suggestions?

kunall17:
What changes you tried to make to things working on the old behaviour file?

harshitagupta30:
I added these 2 lines in `MessageListFragment
`CoordinatorLayout.LayoutParams params = (CoordinatorLayout.LayoutParams) floatingActionButton.getLayoutParams();`
`params.setBehavior(new ShrinkBehaviorOfFab());`

after you said that the behavior is not set.

kunall17:
No, i'm talking about changes in https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/util/RemoveViewsOnScroll.java

harshitagupta30:
I've made no changes in this file instead I've added a new class ShrinkBehaviorOfFab and point the fab behavior to this class

kunall17:
So creating a new behaviour removes the old behaviour as well, which are to supported now as well, so you need to edit the old behaviour!

harshitagupta30:
No, I haven't changed the old behavior.  
This(https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/util/RemoveViewsOnScroll.java) class is having behavior for both app layout and floating button so I added  a new class which are keeping track of both old and new behavior of the fab

kunall17:
So that's what I'm saying so far you need to edit this file https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/util/RemoveViewsOnScroll.java use conditions and sync up with the snackbar/coordinator layout!
So as to have the new beviour and the old one's as well!

harshitagupta30:
I guess the issue may be caused because of using two behaviors coz in `ZulipActivity` we are using 
`RemoveViewsOnScroll` and in `MessageListFragment` we are using `ShrinkBehaviorOfFab` which might be the reason. Let me know if I am thinking right.

harshitagupta30:
Yeah I got it know! I'll make those changes and update the PR. One more thing instead of updating `RemoveViewsOnScroll.java` can I update all the changes related to fab in  `ShrinkBehaviorOfFab.java` and keep `RemoveViewsOnScroll.java` for shrinking AppBarLayout?

kunall17:
Your work will increase way more then, It would be better to have the RemoveViewsOnScroll!

harshitagupta30:
Hi @kunall17! I tried to merge the changes in `RemoveViewsOnScroll` but it is not producing the desired output! So I just wrote the methods `hideView()` and `showView()` in another file and its working fine. Can you please review it!
Thanks :+1:

harshitagupta30:
@kunall17 Can you please review it? :+1:

kunall17:
@harshitagupta30 
The custom behaviour you made is nice, but the problem is the **perfect** sync which was there between the toolbar and the FAB has gone now 
screenshot_20170305-110243

harshitagupta30:
@kunall17 Any suggestions to make it sync again? :)

kunall17:
Yeah use the same behaviour in both AppBarLayout and FAB!

PR #359, opened 2017-02-01
Listen for update message events.

Sam1301:
It'd be nice to see the updated messages in real-time as the user scrolls through the message list.
Added `UpdateMessageWrapper.java` for it.

### DIFF
--- a/app/src/main/java/com/zulip/android/models/Message.java
+++ b/app/src/main/java/com/zulip/android/models/Message.java
@@ -689,10 +689,13 @@ public class Message {
         editDate = history.date;
     }
 
-    public boolean isHasBeenEdited() {
+    public Boolean isHasBeenEdited() {
         return hasBeenEdited;
     }
 
+    public void setHasBeenEdited(boolean isEdited) {
+        this.hasBeenEdited = isEdited;
+    }
 
     //endregion
 
--- a/app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
+++ b/app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
@@ -461,8 +461,9 @@ public class AsyncGetEvents extends Thread {
     }
 
     /**
-     * TODO: add description
-     * @param messageEditLimitEvents
+     * Get updated edit time limit from event of type {@link EventsBranch.BranchType#EDIT_MESSAGE_TIME_LIMIT}
+     *
+     * @param messageEditLimitEvents list of events {@link EventsBranch}
      */
     private void processMessageEditParam(List<EventsBranch> messageEditLimitEvents) {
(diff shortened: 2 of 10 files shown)


smarx:

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -459,4 +474,36 @@ private void processMessageEditParam(List<EventsBranch> messageEditLimitEvents)
...
+                    ZLog.logException(e);
+                }
+            }
+        }
+
+        mActivity.runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+                // notify adapter dataset changed
+                mActivity.getCurrentMessageList().getAdapter().notifyDataSetChanged();
```
kunall17: This notifyDataSetChanged() can be changed as we have the message ID therefore we can find the index of the message in the list and then notifyItemChanged(int);

Sam1301:
Update: added user facing "EDITED" tags in messagetile

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
Sam1301: Done :)

app/src/main/java/com/zulip/android/models/Message.java
```diff
@@ -689,10 +689,13 @@ public void updateFromHistory(@NonNull MessageHistory history) {
         editDate = history.date;
     }
 
-    public boolean isHasBeenEdited() {
```
kunall17: Why this has been changed to `Boolean`?

kunall17:
Also it would be nice to display (DELETED) if the message is deleted?

app/src/main/java/com/zulip/android/models/Message.java
Sam1301: I changed it since [hasBeenEdited is of type Boolean](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/models/Message.java#L120) so on calling this method when hasBeenEdited is null, we would get a null pointer exception `Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference`

kunall17:
After a conversation with @Sam1301 

> Also it would be nice to display (DELETED) if the message is deleted?

This wouldn't look good as already there will be a (Deleted) in the content of the message! 
Rest the PR is awesome! :) 👍 
@timabbott LGTM!

app/src/main/java/com/zulip/android/models/Message.java
kunall17: Ok makes sense! 👍

timabbott:
Awesome.  Merged, thanks @Sam1301!

PR #361, opened 2017-02-02
Fix unnecessary rotation in PhotoSendActivity.

Sam1301:
When a picture is clicked and sent directly (without a crop or edit), it
gets rotated by 90 as most phone cameras are landscape. To solve this,
it'd be better to use the displayed image and save it in the current
photo path before sending.


smarx:

kunall17:
@timabbott LGTM!:)

timabbott:
Merged, thanks @Sam1301!

Issue #343, opened 2017-01-29
Save the server URL after logout


kunall17:
On logging out the first enter server appear's it would be nice to put the server url back there!

harshitagupta30:
@kunall17 Can I work on this issue? :)

kunall17:
Sure go ahead!

kunall17:
Fixed via #363

PR #363, opened 2017-02-03
Issue #343 Save the server URL after Logout resolved

harshitagupta30:
Fixes Issue #343 
@kunall17 Please review! next time I'll make sure of not creating multiple PR! :+1:

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/LoginActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/LoginActivity.java
@@ -41,6 +41,7 @@ import com.zulip.android.networking.response.LoginResponse;
 import com.zulip.android.networking.response.ZulipBackendResponse;
 import com.zulip.android.networking.util.DefaultCallback;
 import com.zulip.android.util.AnimationHelper;
+import com.zulip.android.util.Constants;
 
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -98,6 +99,10 @@ public class LoginActivity extends BaseActivity implements View.OnClickListener,
         mPassword = (EditText) findViewById(R.id.password);
         mShowPassword = (ImageView) findViewById(R.id.showPassword);
         serverIn = (EditText) findViewById(R.id.server_url_in);
+        String serverUrl = getIntent().getStringExtra(Constants.SERVER_URL);
+        if (serverUrl != null) {
+            serverIn.setText(serverUrl);
+        }
         findViewById(R.id.server_btn).setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -100,6 +100,7 @@ import com.zulip.android.networking.AsyncStatusUpdate;
 import com.zulip.android.networking.ZulipAsyncPushTask;
 import com.zulip.android.networking.response.UploadResponse;
 import com.zulip.android.util.AnimationHelper;
+import com.zulip.android.util.Constants;
 import com.zulip.android.util.FilePathHelper;
 import com.zulip.android.util.MutedTopics;
 import com.zulip.android.util.SwipeRemoveLinearLayout;
@@ -306,7 +307,7 @@ public class ZulipActivity extends BaseActivity implements
         processParams();
 
         if (!app.isLoggedIn()) {
-            openLogin();
+            openLogin(null);
             return;
         }
 
@@ -2082,11 +2083,12 @@ public class ZulipActivity extends BaseActivity implements
      */
     private void logout() {
         this.logged_in = false;
+        final String serverUrl = app.getServerURI();
 
         notifications.logOut(new Runnable() {
             public void run() {
                 app.logOut();
-                openLogin();
+                openLogin(serverUrl);
             }
         });
     }
@@ -2094,8 +2096,9 @@ public class ZulipActivity extends BaseActivity implements
     /**
      * Switch to the login view.
      */
-    private void openLogin() {
+    private void openLogin(String serverUrl) {
         Intent i = new Intent(this, LoginActivity.class);
+        i.putExtra(Constants.SERVER_URL, serverUrl);
         startActivity(i);
         finish();
     }
--- a/app/src/main/java/com/zulip/android/util/Constants.java
+++ b/app/src/main/java/com/zulip/android/util/Constants.java
@@ -13,5 +13,6 @@ public class Constants {
     //Default maximum time limit for editing message(Same as server)
     public final static int DEFAULT_MAXIMUM_CONTENT_EDIT_LIMIT = 600;
     public final static boolean DEFAULT_EDITING_ALLOWED = true;
+    public final static String SERVER_URL = "SERVER_URL";
 
 }


smarx:

@harshitagupta30, it looks like you've already signed the Dropbox CLA. Thanks!

app/src/main/java/com/zulip/android/activities/LoginActivity.java
```diff
@@ -98,6 +98,10 @@ protected void onCreate(Bundle savedInstanceState) {
         mPassword = (EditText) findViewById(R.id.password);
         mShowPassword = (ImageView) findViewById(R.id.showPassword);
         serverIn = (EditText) findViewById(R.id.server_url_in);
+        String url = getIntent().getStringExtra("server_url");
```
kunall17: `server_url` declare a constant in constants.java

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -202,6 +202,8 @@ public void onReceive(Context contenxt, Intent intent) {
     private ImageView ivSearchStreamCancel;
     private ListView peopleDrawer;
     private  Toast toast;
+    private String server_url;
```
kunall17: Use camel case here!

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -2096,6 +2099,7 @@ public void run() {
      */
     private void openLogin() {
         Intent i = new Intent(this, LoginActivity.class);
+        i.putExtra("server_url",server_url);
```
kunall17: Format your code!

harshitagupta30:
@kunall17 I have made the changes! Please review!

app/src/main/java/com/zulip/android/activities/LoginActivity.java
```diff
@@ -98,6 +98,10 @@ protected void onCreate(Bundle savedInstanceState) {
         mPassword = (EditText) findViewById(R.id.password);
         mShowPassword = (ImageView) findViewById(R.id.showPassword);
         serverIn = (EditText) findViewById(R.id.server_url_in);
+        String serverUrl = getIntent().getStringExtra("serverUrl");
+        if (serverUrl!=null){
```
yadav-rahul: @harshitagupta30 Formatting needed here.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -2096,6 +2098,7 @@ public void run() {
      */
     private void openLogin() {
         Intent i = new Intent(this, LoginActivity.class);
+        i.putExtra("serverUrl", Constants.serverUrl);
```
vishwesh3: I think it will be better to add all key's in separate class (may be in `Constant.java`)

app/src/main/java/com/zulip/android/util/Constants.java
```diff
@@ -13,5 +13,6 @@
     //Default maximum time limit for editing message(Same as server)
     public final static int DEFAULT_MAXIMUM_CONTENT_EDIT_LIMIT = 600;
     public final static boolean DEFAULT_EDITING_ALLOWED = true;
+    public static String serverUrl = null;
```
vishwesh3: We are using `Constant.java` for all constant used in the project (all are final, no need to set at the runtime).
I think it will be better to use `AppPreferences` for storing `serverUrl`

harshitagupta30: If we declare it as final we can't assign it a new value and it needs to be initialized with some value. I can remove null as of now! Secondly I told in App preferences but @kunall17 told me to use Constants.java

vishwesh3: Ok, no problem

vishwesh3:
Just a suggestion: to achieve mentioned issue we can 
1 ) remove [this](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/ZulipApp.java#L399) line.
2 ) and after [this](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/LoginActivity.java#L100) check if url is available or not.

`serverIn.setText(getApp().getSettings().getString("server_url",""));`

this approach will also cover the case when user log out and completely close the app and after some time launch app(server url will be set).
 
@kunall17 your thoughts on this.

app/src/main/java/com/zulip/android/util/Constants.java
kunall17: Here https://github.com/zulip/zulip-android/pull/363#discussion_r99458136 I asked for the String tag you are passing, not the URL variable!  (text in the bold italics) 
String url = getIntent().getStringExtra("**_server_url_**");

kunall17:
@vishwesh3 Yeah that would work, But it would be better if the logout code is at one place only and it doesn't matter much to pass one string!

app/src/main/java/com/zulip/android/util/Constants.java
harshitagupta30: My bad! I have made changes please review it!

kunall17: Something like this

`public final static String SERVER_URL = "SERVER_URL";`

and use this ` i.putExtra(SERVER_URL, app.getServerURI());` and
`String serverUrl = getIntent().getStringExtra(SERVER_URL);`

harshitagupta30: I guess I can't use _app.getServerURI()_ here because we need to store the url before it get flushed out. So an extra variable will be required.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -2082,20 +2083,22 @@ private void setNightMode(@AppCompatDelegate.NightMode int nightMode) {
      */
     private void logout() {
         this.logged_in = false;
+        final String serverUrl= app.getServerURI();
```
kunall17: Code formatting required here!

harshitagupta30: made changes! :+1:

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -202,7 +203,7 @@ public void onReceive(Context contenxt, Intent intent) {
     private ImageView ivSearchStreamCancel;
     private ListView peopleDrawer;
     private  Toast toast;
-    //
```
kunall17: Remove this change!

harshitagupta30:
@kunall17 I have made the changes! :+1:

kunall17:
And also at here https://github.com/zulip/zulip-android/pull/363/commits/dfc233516a531240ab0b6987ba63c5dadcdf980f#diff-93e07d12c8f17453899c8a8cceb659d2R310 you are passing empty string "" so serverURL here https://github.com/zulip/zulip-android/pull/363/commits/dfc233516a531240ab0b6987ba63c5dadcdf980f#diff-c873fb8dca5a937f04b9a845b2ddd2d8R102 has value "" 

So this can cause problems in the future therefore either check serverUrl is not empty here https://github.com/zulip/zulip-android/pull/363/commits/dfc233516a531240ab0b6987ba63c5dadcdf980f#diff-c873fb8dca5a937f04b9a845b2ddd2d8R103 

or pass null to openLogin() at line 310

harshitagupta30:
@kunall17 I have pass null to the openLogin() in line 310. I hope it will not cause any issue! Please review!

kunall17:
Merged, thanks @harshitagupta30 :)

PR #378, opened 2017-02-09
Change color of send status indicator

### LINKED ISSUE #376: Background color of compose status in night theme is white.
The background and text color of compose status in chatbox is white under night theme.

screenshot_20170210-005751

nitish1211:
Fixes #376

The following colours now show in the sending status indicator.

**Night theme:**
screenshot_20170210-035155

**Day Theme:**

screenshot_20170210-035259

### DIFF
--- a/app/src/main/res/layout/chatbox.xml
+++ b/app/src/main/res/layout/chatbox.xml
@@ -83,16 +83,18 @@
     <LinearLayout
         android:id="@+id/composeStatus"
         android:layout_width="match_parent"
-        android:layout_height="24dp"
-
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
         android:layout_gravity="center"
-        android:background="#eeeeee"
+        android:background="@color/colorPrimary"
         android:gravity="center|center_vertical"
         android:visibility="gone">
 
         <ProgressBar
             android:layout_width="15dp"
-            android:layout_height="15dp" />
+            android:layout_height="15dp"
+            style="@style/AppTheme.DayNight"
+            />
 
         <TextView
             android:id="@+id/sending_message"
@@ -100,6 +102,7 @@
             android:layout_height="wrap_content"
             android:paddingLeft="8dp"
             android:text="@string/sending_message"
+            android:textColor="@color/colorTextPrimary"
             android:textSize="16sp" />
     </LinearLayout>
 
--- a/app/src/main/res/layout/chatbox.xml
+++ b/app/src/main/res/layout/chatbox.xml
@@ -38,7 +38,8 @@
             android:layout_height="wrap_content"
             android:contentDescription="@string/stream_private_content_desp"
             android:padding="3dp"
-            android:src="@drawable/ic_action_person" />
+            android:src="@drawable/ic_action_person"
+            android:tint="@color/colorTextSecondary" />
     </LinearLayout>
 
     <LinearLayout
@@ -51,13 +52,14 @@
             android:id="@+id/camera_btn"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
+            android:alpha="0.7"
             android:contentDescription="@string/camera_content_desp"
             android:paddingBottom="4dp"
             android:paddingLeft="8dp"
             android:paddingRight="4dp"
             android:paddingTop="4dp"
             android:src="@drawable/ic_photo_camera_black_24dp"
-            android:tint="#757575" />
+            android:tint="@color/colorTextSecondary" />
 
         <AutoCompleteTextView
             android:id="@+id/message_et"
@@ -73,10 +75,11 @@
             android:id="@+id/send_btn"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
+            android:alpha="0.7"
             android:contentDescription="@string/send_content_desp"
             android:padding="4dp"
             android:src="@drawable/ic_send_24dp"
-            android:tint="#757575" />
+            android:tint="@color/colorTextSecondary" />
 
     </LinearLayout>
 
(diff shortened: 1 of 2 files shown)


smarx:

@nitish1211, it looks like you've already signed the Dropbox CLA. Thanks!

Sam1301:
Thanks for working on this :+1: , just a UI preference, I think it would be better if we could match the background color of compose status to the chatbox color in both themes.

nitish1211:
@Sam1301 Done.

Sam1301:
Nice, works as expected in day theme. For night theme, I think the background color of the compose status could be made to match the color of the chatbox since we are doing _the same in the day theme_

current
screenshot_20170213-050329

expected
screenshot_20170213-052246

kunall17:
This seems much better and required, also how about changing the chatBox text colors and the timestamp color, maybe 2 shades lighter?

kunall17:
The text color seems fines, I was judging from the screenshots you posted but forget that they are while we are sending messages!

Btw can you fix these two bugs in different commits in this PR itself?

- Different icon colors in the chat box
- In the stream textView in the message header that triangle thing should be coloured dark!

screen shot 2017-02-14 at 11 27 31 pm
screen shot 2017-02-14 at 11 28 53 pm

nitish1211:
@kunall17 Is there a specific reason why some features are excluded from the night theme?
In these cases:
Prevents setting of the color of the arrow : [Line 357 ](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L357)

Setting of the color bar: [Line 409](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L409)

kunall17:
>  why some features are excluded from the night theme?

Didn't understand you correctly?

nitish1211:
`if (!isCurrentThemeNight)` [here](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L357)  why has this condition been used before setting color of the arrow(triangular thing) and when setting the color of the color bar( which is the  colored bar you would notice enclosing the messages under a particular stream on the left edge of the screen).

I was just wondering if there was any design specific reason why this was done.

nitish1211:
Here I've made changes to the Night theme to make it slightly similar to the day theme.
1)The leftBar is now visible
2)The arrow has the same color as the label.

Before:
screenshot_20170216-014345


After:
screenshot_20170216-015830

app/src/main/res/layout/chatbox.xml
```diff
@@ -76,25 +77,24 @@
...
 
     <LinearLayout
         android:id="@+id/composeStatus"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:orientation="horizontal"
         android:layout_gravity="center"
         android:background="@color/colorPrimary"
         android:gravity="center|center_vertical"
+        android:orientation="horizontal"
```
Sam1301: nit: only ordering of attributes changed

app/src/main/res/layout/chatbox.xml
```diff
@@ -76,25 +77,24 @@
...
         android:visibility="gone">
 
         <ProgressBar
-            android:layout_width="15dp"
-            android:layout_height="15dp"
             style="@style/AppTheme.DayNight"
-            />
+            android:layout_width="15dp"
+            android:layout_height="15dp" />
```
Sam1301: same as above

kunall17:
Also it would be better if you split up the fixing of message header colors!

app/src/main/res/layout/chatbox.xml
nitish1211: This happens on **formatting** the code. Android studio sorts the attributes in a particular order.

Sam1301: ohk.. ya right :+1: , its fine then :)

nitish1211:
@kunall17 Done. Added a commit for color in the header and side bar in the night theme.

kunall17:
These left color bars were not added on purpose, minimal colors are to be included in the dark theme, remove them! and it's better to have the old format (using isCurrentThemeNight)

nitish1211:
So you prefer color in header arrows but not in sidebar or should it be left as it was earlier? @kunall17 

Like this :
screenshot_20170215-210738

kunall17:
Yeah!

nitish1211:
@kunall17 Done. Added the changes you requested. :+1:

kunall17:
Looks nice, but the color tint on the person icon is not the same color with the other colors! 

screen shot 2017-02-22 at 9 38 23 pm

kunall17:
Also on the day theme this result is being seen! 
screen shot 2017-02-22 at 10 02 22 pm

nitish1211:
@kunall17 This is happening due to the difference in color between the icon's default color which is not dark enough. 

This is the person icon:

ic_action_person

These are the other icons:

ic_photo_camera_black_24dp

ic_send_24dp


Which are darker by default causing them to appear darker. To change the darkness there are two options either to change the icon file or add another color to the person icon which is slightly darker so that both of them look the same. Which one would you prefer?

nitish1211:
Btw this is how it looks on making all icons the same by making the others lighter

screenshot_20170223-210605

screenshot_20170223-210613

Is this fine?

kunall17:
No in the light theme it should be black icons only!

nitish1211:
screenshot_20170223-231607

screenshot_20170223-231559

How about this?

kunall17:
Merged, thanks @nitish1211 👍

PR #390, opened 2017-02-10
Fixes #386 : Compose box doesn't auto-open when narrowing

### LINKED ISSUE #386: Compose box shouldn't auto-open when narrowing
When you click on a message header to narrow to a stream/topic pair, the compose box opens automatically, which can be distracting and takes up space.  We probably don't want it to open unless the user has done something that indicates they actually want to compose a reply (such as tapping in the message body rather than the recipient headers).

saketkumar95:
**Summary of changes**

This PR fixes https://github.com/zulip/zulip-android/issues/386 
Now compose box doesn't auto-open when narrowing.

chatbox

- [x] Code is formatted.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -1204,6 +1204,8 @@ public class ZulipActivity extends BaseActivity implements
                         String subjectName = ((Cursor) streamsDrawer.getExpandableListAdapter().getChild(groupPosition, childPosition)).getString(0);
                         onNarrow(new NarrowFilterStream(streamName, subjectName));
                         onNarrowFillSendBoxStream(streamName, subjectName, false);
+                        hideView(chatBox);
+                        displayFAB(true);
                         break;
                     default:
                         return false;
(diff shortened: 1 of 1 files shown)


smarx:

@saketkumar95, it looks like you've already signed the Dropbox CLA. Thanks!

timabbott:
Yes, the issue was with message headers ("recipient bars" is what we call them usually).

saketkumar95:
@timabbott @Sam1301 Yeah, i got that wrong. But i think this should also be there. Whenever someone clicks on the stream, the compose box automatically opens up. Instead it should not. Same as on web.
Thoughts?

timabbott:
Agreed.

saketkumar95:
@timabbott Yeah, sure. :)

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -1204,6 +1204,8 @@ public boolean onChildClick(ExpandableListView parent, View v, int groupPosition
                         String subjectName = ((Cursor) streamsDrawer.getExpandableListAdapter().getChild(groupPosition, childPosition)).getString(0);
                         onNarrow(new NarrowFilterStream(streamName, subjectName));
                         onNarrowFillSendBoxStream(streamName, subjectName, false);
+                        hideView(chatBox);
```
Sam1301: would be better to use displayChatbox(boolean) and displayFAB(boolean)

Sam1301:
@saketkumar95 this looks nice, can you please address the comment here?

PR #408, opened 2017-02-18
Fetch raw message content before editing

kunall17:
**Summary of changes**

[Explain here what changes were made and why.]

This now fetches the raw message before editing to save the formatted content (like emojis, mention @) 

Still one thing can be noticed after edit is successful, the updated raw content is displayed instead of the formatted content, but after list reload it start's showing fine. 

I'll update the PR if I find solution for this! 

- [x] Code is formatted.

- [x]  Commit messages are well-written.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
+++ b/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
@@ -234,7 +234,6 @@ public class MessageListFragment extends Fragment implements MessageListener {
                 return true;
             case R.id.copy_message:
                 copyMessage(message);
-                Toast.makeText(getContext(), R.string.message_copied, Toast.LENGTH_SHORT).show();
                 return true;
             case R.id.edit_message:
                 editMessage(message, adapter.getContextMenuItemSelectedPosition());
@@ -510,17 +509,49 @@ public class MessageListFragment extends Fragment implements MessageListener {
 
     @SuppressWarnings("deprecation")
     @TargetApi(Build.VERSION_CODES.HONEYCOMB)
-    private void copyMessage(Message msg) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
-            android.content.ClipboardManager clipboard = (android.content.ClipboardManager) app
-                    .getSystemService(Context.CLIPBOARD_SERVICE);
-            clipboard.setPrimaryClip(ClipData.newPlainText("Zulip Message",
-                    msg.getContent()));
-        } else {
-            android.text.ClipboardManager clipboard = (android.text.ClipboardManager) app
-                    .getSystemService(Context.CLIPBOARD_SERVICE);
-            clipboard.setText(msg.getContent());
-        }
+    private void copyMessage(final Message message) {
+        final CommonProgressDialog commonProgressDialog = new CommonProgressDialog(getContext());
+        commonProgressDialog.showWithMessage(getString(R.string.copy_message_fetch));
+        app.getZulipServices()
+                .fetchRawMessage(message.getID())
+                .enqueue(new DefaultCallback<RawMessageResponse>() {
+                    @Override
+                    public void onSuccess(Call<RawMessageResponse> call, Response<RawMessageResponse> response) {
+                        RawMessageResponse messageResponse = response.body();
+                        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.HONEYCOMB) {
+                            android.content.ClipboardManager clipboard = (android.content.ClipboardManager) app
+                                    .getSystemService(Context.CLIPBOARD_SERVICE);
+                            clipboard.setPrimaryClip(ClipData.newPlainText("Zulip Message",
+                                    messageResponse.getRawContent()));
+                        } else {
+                            android.text.ClipboardManager clipboard = (android.text.ClipboardManager) app
+                                    .getSystemService(Context.CLIPBOARD_SERVICE);
+                            clipboard.setText(messageResponse.getRawContent());
+                        }
+                        Toast.makeText(getActivity(), R.string.message_copied, Toast.LENGTH_SHORT).show();
+                        commonProgressDialog.dismiss();
+                    }
+
+                    @Override
+                    public void onError(Call<RawMessageResponse> call, Response<RawMessageResponse> response) {
+                        RawMessageResponse messageResponse = response.body();
+                        if (messageResponse != null) {
+                            Toast.makeText(getActivity(), (TextUtils.isEmpty(messageResponse.getMsg())) ? getString(R.string.copy_failed) :
+                                    messageResponse.getMsg(), Toast.LENGTH_SHORT).show();
+                        } else {
+                            Toast.makeText(getActivity(), R.string.copy_failed, Toast.LENGTH_SHORT).show();
+                        }
+                        //If something fails msg will have something to display
+                        commonProgressDialog.dismiss();
+                    }
+
+                    @Override
+                    public void onFailure(Call<RawMessageResponse> call, Throwable t) {
+                        super.onFailure(call, t);
+                        commonProgressDialog.dismiss();
+                        Toast.makeText(getActivity(), R.string.copy_failed, Toast.LENGTH_SHORT).show();
+                    }
+                });
     }
 
     public void onNewMessages(Message[] messages) {
--- a/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
+++ b/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
@@ -13,6 +13,7 @@ import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
+import android.text.TextUtils;
 import android.util.Log;
 import android.util.SparseArray;
 import android.view.LayoutInflater;
@@ -39,6 +40,9 @@ import com.zulip.android.models.Stream;
 import com.zulip.android.networking.AsyncGetOldMessages;
 import com.zulip.android.networking.ZulipAsyncPushTask;
 import com.zulip.android.networking.response.EditResponse;
+import com.zulip.android.networking.response.RawMessageResponse;
+import com.zulip.android.networking.util.DefaultCallback;
+import com.zulip.android.util.CommonProgressDialog;
 import com.zulip.android.util.Constants;
 import com.zulip.android.util.MessageListener;
 import com.zulip.android.util.MutedTopics;
@@ -174,7 +178,7 @@ public class MessageListFragment extends Fragment implements MessageListener {
                 }
             }
         });
-        mListener.setLayoutBehaviour(linearLayoutManager,adapter);
+        mListener.setLayoutBehaviour(linearLayoutManager, adapter);
         return view;
     }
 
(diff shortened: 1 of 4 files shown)


smarx:

kunall17:
@yadav-rahul If you had done work for this one, and it got wasted then sorry the release is being pending for weeks and you have not responded! 

@Sam1301 Review!

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -398,70 +401,101 @@ private void editMessage(final Message message, final int position) {
...
-                                }
+            final CommonProgressDialog commonProgressDialog = new CommonProgressDialog(getContext());
+            commonProgressDialog.showWithMessage(getString(R.string.fetch_edit_message));
+            app.getZulipServices()
+                    .fetchRawMessage(message.getID())
+                    .enqueue(new Callback<RawMessageResponse>() {
+                        @Override
+                        public void onResponse(Call<RawMessageResponse> call, Response<RawMessageResponse> response) {
+                            RawMessageResponse messageResponse = response.body();
+                            if (!messageResponse.getResult().equalsIgnoreCase("success")) {
```
Sam1301: response.isSuccessful() can be used here; even better would be to use [DefaultCallback abstract class](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/networking/util/DefaultCallback.java)

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -398,70 +401,101 @@ private void editMessage(final Message message, final int position) {
...
+                                Toast.makeText(getContext(), (messageResponse.getMsg() == null || TextUtils.isEmpty(messageResponse.getMsg())) ? getString(R.string.message_edit_failed) :
```
Sam1301: I think textUtils already checks for str == null, hence we can omit `messageResponse.getMsg() == null || `

app/src/main/res/values/strings.xml
```diff
@@ -151,4 +151,5 @@
     <string name="connecting_to_server">Connecting to server</string>
     <string name="refreshing">Refreshing</string>
     <string name="logging_out">Logging out</string>
+    <string name="fetch_edit_message">Preparing message for Edit</string>
```
Sam1301: we might consider using "Preparing message for Edit" which better matches the zulip capitalization scheme?

Sam1301:
@kunall17 reviewed :+1: Also just noticed something similar happens for coping a message... the formatted content is lost.

Sam1301:
I am facing one issue with this. When I edit a message with any type of formatting, and press edit without changing any text, only raw message is visible and markdown isn't render.
g_20170219_0936563

Sam1301:
oh.. just noticed your comment "Still one thing can be noticed after edit is successful, the updated raw content is displayed instead of the formatted content, but after list reload it start's showing fine.", I guess then its fine :) :+1:

kunall17:
Done 👍 
Also added a check if no changes then show a toast!

kunall17:
Pushed a commit to fetch raw content on copy to clipboard as well!

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -398,70 +402,102 @@ private void editMessage(final Message message, final int position) {
...
-                                        message.setFormattedContent(editedMessageContent);
-                                        adapter.notifyItemChanged(position);
-                                        progress.dismiss();
-                                        Toast.makeText(getActivity(), R.string.message_edited, Toast.LENGTH_SHORT).show();
-                                    } else {
-                                        progress.dismiss();
-                                        Toast.makeText(getActivity(), R.string.message_edit_failed, Toast.LENGTH_SHORT).show();
-                                    }
-                                }
+            final CommonProgressDialog commonProgressDialog = new CommonProgressDialog(getContext());
```
Sam1301: I think we might wanna override onFailure() to dismiss CommonProgressDialog as well since DefaultCallback doesn't take care of that (https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/networking/util/DefaultCallback.java#L30).

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -500,17 +499,39 @@ public void onFailure(Call<EditResponse> call, Throwable t) {
...
-                    .getSystemService(Context.CLIPBOARD_SERVICE);
-            clipboard.setPrimaryClip(ClipData.newPlainText("Zulip Message",
-                    msg.getContent()));
-        } else {
-            android.text.ClipboardManager clipboard = (android.text.ClipboardManager) app
-                    .getSystemService(Context.CLIPBOARD_SERVICE);
-            clipboard.setText(msg.getContent());
-        }
+    private void copyMessage(final Message message) {
+        final CommonProgressDialog commonProgressDialog = new CommonProgressDialog(getContext());
```
Sam1301: (same comment) I think we might wanna override onFailure() to dismiss CommonProgressDialog as well since DefaultCallback doesn't take care of that (https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/networking/util/DefaultCallback.java#L30).

app/src/main/res/values/strings.xml
```diff
@@ -153,4 +153,6 @@
     <string name="logging_out">Logging out</string>
     <string name="fetch_edit_message">Preparing Message for Edit</string>
     <string name="no_edit">No Edits Detected</string>
+    <string name="copy_message_fetch">Copying Message To Clipboard</string>
+    <string name="copy_failed">Copying to Clipboard Failed</string>
```
Sam1301: same comment (#409)

kunall17:
Done :)

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -398,83 +401,148 @@ private void editMessage(final Message message, final int position) {
...
+                        public void onSuccess(Call<RawMessageResponse> call, Response<RawMessageResponse> response) {
+                            RawMessageResponse messageResponse = response.body();
+                            commonProgressDialog.dismiss();
+                            showEditMessageDialog(message, messageResponse.getRawContent(), position);
+                        }
+
+                        @Override
+                        public void onError(Call<RawMessageResponse> call, Response<RawMessageResponse> response) {
+                            RawMessageResponse messageResponse = response.body();
+                            Toast.makeText(getActivity(), (TextUtils.isEmpty(messageResponse.getMsg())) ? getString(R.string.message_edit_failed) :
```
Sam1301: We might wanna have a check here if messageResponse is null; for eg: in case of an internal server error, the response is null.
I get the following crash we the onError is called
```
E/AndroidRuntime: FATAL EXCEPTION: main
                                                                       Process: com.zulip.android.dev, PID: 30231
                                                                       java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.zulip.android.networking.response.RawMessageResponse.getMsg()' on a null object reference
                                                                           at com.zulip.android.activities.MessageListFragment$6.onError(MessageListFragment.java:528)
                                                                           at com.zulip.android.networking.util.DefaultCallback.onResponse(DefaultCallback.java:20)
                                                                           at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:68)
                                                                           at android.os.Handler.handleCallback(Handler.java:739)
                                                                           at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                           at android.os.Looper.loop(Looper.java:148)
                                                                           at android.app.ActivityThread.main(ActivityThread.java:7325)
                                                                           at java.lang.reflect.Method.invoke(Native Method)
                                                                           at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
                                                                           at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
```

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -398,83 +401,148 @@ private void editMessage(final Message message, final int position) {
...
+                                    .getSystemService(Context.CLIPBOARD_SERVICE);
+                            clipboard.setText(messageResponse.getRawContent());
+                        }
+                        Toast.makeText(getActivity(), R.string.message_copied, Toast.LENGTH_SHORT).show();
+                        commonProgressDialog.dismiss();
+                    }
+
+                    @Override
+                    public void onError(Call<RawMessageResponse> call, Response<RawMessageResponse> response) {
+                        RawMessageResponse messageResponse = response.body();
```
Sam1301: same comment as above

app/src/main/res/values/strings.xml
```diff
@@ -151,4 +151,8 @@
     <string name="connecting_to_server">Connecting to server</string>
     <string name="refreshing">Refreshing</string>
     <string name="logging_out">Logging out</string>
+    <string name="fetch_edit_message">Preparing Message for edit</string>
+    <string name="no_edit">No edits detected</string>
+    <string name="copy_message_fetch">Copying message To clipboard</string>
```
Sam1301: nit: To -> to :)

Sam1301:
@kunall17 Nice work :) :+1: , should we merge this before fixing the "raw content being displayed" bug?

kunall17:
@Sam1301 what do you think?

Sam1301:
So currently if message is edited with formatting, the user can't see the formatted content back. Its better than the previous version since now some meaningful text is displayed. The user might not press refresh to see the content. I think UI wise it would make sense to solve the bug otherwise maybe not merge it now. I'd like to know your views on this :).

Sam1301:
I think the copy message fix can definitely be merge now :+1:

kunall17:
Updated with the fix! :)

ezgif-1-5c00886281

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -507,9 +509,17 @@ private void processUpdateMessages(List<EventsBranch> updateEvents) {
...
+                int i = 0;
                 for (int id : messageIds) {
                     // notify adapter data item changed
-                    adapter.notifyItemChanged(adapter.getItemIndex(id));
+                    int index = adapter.getItemIndex(id);
+                    adapter.notifyItemChanged(index);
+                    RecyclerView.ViewHolder viewHolderForAdapterPosition = mActivity.getCurrentMessageList().getRecyclerView().findViewHolderForAdapterPosition(index);
+                    if (viewHolderForAdapterPosition instanceof MessageHolder) {
+                        UpdateMessageWrapper updateEvent = (UpdateMessageWrapper) updateEvents.get(i++);
+                        Spanned spannedString = Message.formatContent(updateEvent.getFormattedContent(), ZulipApp.get());
```
Sam1301: [app](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java#L62) can be used

app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -398,83 +401,156 @@ private void editMessage(final Message message, final int position) {
...
+                //Start a progress dialog indicating editing message
+                final ProgressDialog progress = new ProgressDialog(getActivity());
+                progress.setCancelable(false);
+                progress.setMessage(app.getString(R.string.editing_message));
+                progress.show();
+                final String editedMessageContent =
+                        dialogMessageEditText.getText().toString().length() == 0 ? getString(R.string.default_delete_text) : dialogMessageEditText.getText().toString();
+
+                if (editedMessageContent.equals(rawContent)) {
+                    Toast.makeText(getActivity(), R.string.no_edit, Toast.LENGTH_SHORT).show();
```
Sam1301: we might wanna dismiss progress dialog here

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
```diff
@@ -507,9 +509,17 @@ private void processUpdateMessages(List<EventsBranch> updateEvents) {
...
                 RecyclerMessageAdapter adapter = mActivity.getCurrentMessageList().getAdapter();
+                int i = 0;
                 for (int id : messageIds) {
                     // notify adapter data item changed
-                    adapter.notifyItemChanged(adapter.getItemIndex(id));
+                    int index = adapter.getItemIndex(id);
+                    adapter.notifyItemChanged(index);
+                    RecyclerView.ViewHolder viewHolderForAdapterPosition = mActivity.getCurrentMessageList().getRecyclerView().findViewHolderForAdapterPosition(index);
+                    if (viewHolderForAdapterPosition instanceof MessageHolder) {
+                        UpdateMessageWrapper updateEvent = (UpdateMessageWrapper) updateEvents.get(i++);
```
Sam1301: I think an assumption is made here that all updateEvents will have a corresponding messageId in the arraylist but I believe that isn't true https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java#L499 for all scenarios.

Sam1301:
@kunall17 I faced one issue with this. When I tried to omit some text, I got the bug again. After this, on refresh, the markdown was rendered correctly.
g_20170220_0410063

kunall17:
Now this is super weird, when I deleted one smiley from two then it rendered correctly, 
If i add something to the list (after editing) it works

If i delete something in the list it doesn't! :/

kunall17:
Well there is something wrong with the list rendering @Sam1301 
If there is a space in the end then the last letter is removed in the rendered text! 

So when there are two something like (no trailing spaces) 
```
* ASD
* ZXC
```
And you edit them something like (underscore representing space)

`* ASD_`

The list is rendered! 

bugs
device-2017-02-20-201613

app/src/main/java/com/zulip/android/networking/AsyncGetEvents.java
kunall17: I have added a check https://github.com/zulip/zulip-android/pull/408/files#diff-12a339ef893658709ac106c1e3a45a5eR516 so if there doesn't exists a message of that ID in the list it will skip that message!

Sam1301:
Hmm weird... I doubt if the problem is with _list rendering_ since on refresh the list is rendered correctly but only after edit the list isn't rendered correctly :/

I tried it with different formatting options : quote block, inline code, code block, etc. but got the same bug for them as well while omitting text.
Here's the gif
g_20170220_2039173

kunall17:
Fixed! 
final

But I can confirm [the list rendering](https://cloud.githubusercontent.com/assets/12700799/23129731/d9455318-f7a9-11e6-8d95-2343bb257e3a.gif) with a space at end isn't right  and it has nothing to do with editing!

Sam1301:
@kunall17 I got the same bug with list rendering: one letter isn't displayed at the end...looking into it :+1: 

For edit bug, works well for quote, list, inline code :+1: I noticed an extra newline added when I edited for these formatting options, is that intentional?
For other formatting options like quote block, emoji, code block, unfortunately I got the same bug... any clue why?
Here's a [gif](https://drive.google.com/open?id=0B4jK5QX65b0ZNWwxOURkdV85ODg) for it

Sam1301:
@kunall17 I found something that might help solve the issue
* the newlines are lost when message is edited
* when notifyItemChanged is called, onBindViewHolder gets called internally to reset the view. Hence, formatted content is set on the textview(https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L380) when update_message event is received. So, I think we don't need a change in AsyncGetEvents to solve this?

kunall17:
> the newlines are lost when message is edited

That's the expected result, same is followed in the web UI

> notifyItemChanged

I know but by manually rendering to the holder, I see better results
Even i'm not able to understand why this bug is occurring!

Sam1301:
I think conceptually we might want to imitate the behaviour of how we process message update events(on master branch): we edit a message with any combination of markdown, the event is received, data set is notified and the message tile is rendered again.
After edit we receive the update message event, so we can leave all the updation work of views to the event.. and maybe do nothing after editing in messagelistfragment

kunall17:
This now fixes the whole thing! 

Thanks to @Sam1301 for providing a better solution! :) 👍

Sam1301:
I think we should now do some final testing, just to be sure everything is fixed :) Then, we'll merge this :+1:

Sam1301:
Awesome, merged thanks @kunall17 ! :) :+1:

PR #415, opened 2017-02-24
Fix: Crash on context menu item selected when fragment replaced.

Sam1301:
When a message with a clickable span is long pressed, the user can accidentally long press on the span itself, which can cause a fragment replace transaction and contextmenu generation at the same time. On selecting any of the options of menu the app crashes. Actually this bug is also in v1.1.2
Added a check for this.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
+++ b/app/src/main/java/com/zulip/android/activities/MessageListFragment.java
@@ -202,7 +202,14 @@ public class MessageListFragment extends Fragment implements MessageListener {
 
     @Override
     public boolean onContextItemSelected(MenuItem item) {
-        Message message = (Message) adapter.getItem(adapter.getContextMenuItemSelectedPosition());
+        // fragment replace transaction can be committed during long press on a message
+        int position = adapter.getContextMenuItemSelectedPosition();
+        if (position == RecyclerView.NO_POSITION) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/zulip/android/activities/MessageListFragment.java
```diff
@@ -202,7 +202,13 @@ public void onDetach() {
 
     @Override
     public boolean onContextItemSelected(MenuItem item) {
-        Message message = (Message) adapter.getItem(adapter.getContextMenuItemSelectedPosition());
+        // fragment replace transaction can be committed during long press on a message
+        int position = adapter.getContextMenuItemSelectedPosition();
+        if (position == RecyclerView.NO_POSITION) {
+            return false;
```
kunall17: It would be better if we would show the user some message as toast as well!

Sam1301:
updated branch

app/src/main/res/values/strings.xml
```diff
@@ -155,4 +155,5 @@
     <string name="no_edit">No edits detected</string>
     <string name="copy_message_fetch">Copying message to clipboard</string>
     <string name="copy_failed">Copying to clipboard failed</string>
+    <string name="message_not_found">Message not found</string>
```
kunall17: Technically this message is right, but you should change this string as the user will kinda annoyed as they would have long press on the message itself! 
maybe something like Something went wrong, try again

Sam1301: makes sense :+1: updated the branch

PR #416, opened 2017-02-24
Fix: Use Patterns.WEB_URL to check if url valid or not.

Sam1301:
So `URLUtil.isValidUrl` isn't the best method to check validity of url since it returns true for addresses like "https://something  and.com".

Used `Patterns.WEB_URL` instead. This bug too has been in v1.1.2

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/LoginActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/LoginActivity.java
@@ -19,7 +19,6 @@ import android.util.Log;
 import android.util.Patterns;
 import android.view.View;
 import android.view.inputmethod.InputMethodManager;
-import android.webkit.URLUtil;
 import android.widget.Button;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/zulip/android/activities/LoginActivity.java
```diff
@@ -279,6 +278,7 @@ private void showBackends(String httpScheme, String serverURL) {
 
         if (!isUrlValid(serverURL)) {
             Toast.makeText(LoginActivity.this, R.string.invalid_url, Toast.LENGTH_SHORT).show();
+            commonProgressDialog.dismiss();
```
kunall17: This change in another commit!

Sam1301:
Updated 👍

kunall17:
Merged, Thanks @Sam1301 👍

PR #430, opened 2017-03-03
Show unread counts and add bankruptcy option.

Sam1301:
**Summary of changes**

* shows unread counts
  * real-time sync with server using update_message_flags event
  * UI changes

* bankruptcy menu option
  * fetch all messages in batches of 200

* fetch 1000 messages on start-up (related discussion https://github.com/zulip/zulip-android/pull/283#issuecomment-273869004)

### DIFF
--- a/app/src/main/java/com/zulip/android/database/DatabaseHelper.java
+++ b/app/src/main/java/com/zulip/android/database/DatabaseHelper.java
@@ -30,7 +30,7 @@ public class DatabaseHelper extends OrmLiteSqliteOpenHelper {
     private static final String DATABASE_NAME = "zulip-%s.db";
     // any time you make changes to your database objects, you may have to
     // increase the database version
-    private static final int DATABASE_VERSION = 8;
+    private static final int DATABASE_VERSION = 9;
 
     public DatabaseHelper(ZulipApp app, String email) {
         super(app, String.format(DATABASE_NAME, MD5Util.md5Hex(email)), null,
--- a/app/src/main/res/layout-v21/stream_tile_child.xml
+++ b/app/src/main/res/layout-v21/stream_tile_child.xml
@@ -41,6 +41,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:background="@drawable/unread_counts_background"
+            android:elevation="2dp"
             android:gravity="center"
             android:textColor="@android:color/white"
             android:text="2"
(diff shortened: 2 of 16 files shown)


timabbott:
So the main potential downside of fetching 1000 messages on startup is performance when on a slow network; to what extent have you tested that?  It might actually be fine with compression, but we should be sure :)

Sam1301:
I tested the difference between fetching 100 and 1000 messages. Here are the results:
For 100 messages (current behaviour)
for100msgs

For 1000 messages
for_1000msgs

these results were obtained for a 35Mbps connection... i think we could fetch 1000 messages in batches of 200 or so and display them simultaneously?

timabbott:
So I think with a 35Mbps connection, it'll be fine either way.  For this, we want to test with a relatively slow connection.  

We could do something simpler than doing 5 batches of 200, like fetch 200 at first (to show something) and then fetch the next 800.  The thing that's really important to optimize here is the time-to-first-interaction, but there's not much reason to do 5 rounds over 2.

Sam1301:
ok got it, I'll update the pr with fetching 200 first and displaying it. Following this, in the background 800 messages can be fetched and added to the bottom of the list. During this time, the user will be able to interact with those 200 messages in the list. These 800 messages loaded in background could probably be fetched in a single call. Would that be fine?

timabbott:
Yep, fetching 800 in one call makes sense.  FWIW the webapp does roughly that with 400 and 1000 as the parameters, so it's a pretty similar algorithm.

app/src/main/java/com/zulip/android/database/DatabaseHelper.java
```diff
@@ -30,7 +30,7 @@
     private static final String DATABASE_NAME = "zulip-%s.db";
     // any time you make changes to your database objects, you may have to
     // increase the database version
-    private static final int DATABASE_VERSION = 8;
+    private static final int DATABASE_VERSION = 9;
```
saketkumar95: Database version is already updated here https://github.com/zulip/zulip-android/commit/b596202484f28639c22ec684f51572ff1767fda7

app/src/main/res/layout-v21/stream_tile_child.xml
```diff
@@ -0,0 +1,50 @@
...
+        android:layout_alignParentLeft="true"
+        android:layout_alignParentStart="true"
+        android:layout_alignParentTop="true"
+        android:layout_gravity="center_vertical"
+        android:layout_toLeftOf="@+id/child_container"
+        android:layout_marginLeft="54dp"
+        android:paddingBottom="16dp"
+        android:paddingLeft="16dp"
+        android:paddingRight="16dp"
+        android:paddingTop="16dp"
```
saketkumar95: It would be better if you put this all in dimens.xml. :)

saketkumar95:
@Sam1301 Really nice work! :) Resolve the merge conflicts too. :+1:

app/src/main/java/com/zulip/android/database/DatabaseHelper.java
Sam1301: @saketkumar95 we need to update the database version whenever we make changes to the existing database model so that the existing tables are dropped and new tables with updated schema are created.

Issue #385, opened 2017-02-10
Add a floating recipient bar


timabbott:
Right now, it can sometimes be pretty hard to tell what stream/topic you're looking at in the home view (or which topic in a stream view) that has a lot of messages in a row to the same topic, because we don't have a floating recipient bar showing the stream/topic (or whatever the recipient is) for the messages at the top of the view.  The webapp does this well, as does the React Native iOS app.

vishwesh3:
I would like to work on this.

Sam1301:
@vishwesh3 go for it :+1: :)

vishwesh3:
Right now if same user send multiple messages continuously then his/her name field is added again along with message. I think it will be better to do something like web

![148865237163881](https://cloud.githubusercontent.com/assets/18511177/23581290/27df9fba-0137-11e7-86a1-13572fc61e25.png)
 

![screenshot_1488651457](https://cloud.githubusercontent.com/assets/18511177/23581218/131043d8-0136-11e7-9452-68b24c2f7a72.png)


@timabbott @kunall17 @Sam1301 Your thoughts?

timabbott:
Yes, definitely.

vibhav16:
i would like to work on this issue.

vishwesh3:
@vibhav16 Thanks for your interest in this.
This is already fixed in https://github.com/zulip/zulip-android/pull/437

PR #437, opened 2017-03-04
Added floating recipient bar.

vishwesh3:
Fix:#385

**Summary of changes**

ezgif com-video-to-gif 9

- [x] Code is formatted.


vishwesh3:
Update

Now Android message list resembles with web message list.

- Floating recipient bar.
- Recipient bar contains date.
- Message list have date separator between messages.
- Hide avatar and sender name if send by same sender.
- Show `edited` tag, star status and time in left for the second message onward (messages send by same user)
- Hide timestamp of messages send in small interval by same sender.
- Message tile contains only time (date in recipient bar and date separator like in web)

**Screenshots**

screenshot_20170314-032110
screenshot_20170314-032118
screenshot_20170314-032130
screenshot_20170314-032213
screenshot_20170314-032332

vishwesh3:
resolved merge conflicts.

kunall17:
@vishwesh3 this looks awesome, just noticed one problem if you scroll towards older messages (going up) then the toolbar is made visible, and the sticky header gets hidden, 

Any solutions to it?

vishwesh3:
@kunall17 One possible solution: - now sticky headers floats below toolbar if it is shown. 

ezgif com-video-to-gif 27

kunall17:
@vishwesh3 

Faced problems like these
This can be fixed! 
[screenshot_20170419-185828](https://cloud.githubusercontent.com/assets/12700799/25182471/4d38b962-2532-11e7-98de-5196f91edf32.png)

vishwesh3:
@kunall17 Updated :+1: 

This is how it look's now.
ezgif com-optimize

app/src/main/java/com/zulip/android/viewholders/stickyheaders/interfaces/RetrieveHeaderView.java
```diff
@@ -0,0 +1,38 @@
...
+import android.support.v7.widget.RecyclerView;
+import android.view.ViewGroup;
+
+public interface RetrieveHeaderView {
+
+    int DEFAULT_VIEW_TYPE = -1;
+
+    RecyclerView.ViewHolder getViewHolderForPosition(int headerPositionToShow);
+
+    class RecyclerRetrieveHeaderView implements RetrieveHeaderView {
```
kunall17: This file class structure can be improved!

kunall17:
Also a little layout issue, the timestamp is not vertically centered! 

device-2017-04-22-001751

app/src/main/java/com/zulip/android/viewholders/stickyheaders/interfaces/RetrieveHeaderView.java
vishwesh3: Done :+1:

kunall17:
Cool, @vishwesh3 there are multiple yesterday, today tags coming up is that intentional or a bug? 

(Even if its a bug, I don't have much problems with it) would like to know thought's of other folks @Sam1301 ?

Sam1301:
> (Even if its a bug, I don't have much problems with it) would like to know thought's of other folks @Sam1301 ?

The floating headers look really nice; I am in favour of merging this. Though I feel like the timestamps are *too small* to read.

Also we might want to wrap the header topics to max 2 lines since on small screen devices, it looks something like this 
whatsapp image 2017-04-22 at 18 54 02

kunall17:
That looks awful, the main problem with this one the stream name is also really long hence even for two lines we won't get much space and end up loosing characters. 

How about we go for ellipsis for the topic textview?

vishwesh3:
if we go with `ellipsize="end" ` and `maxLines="1"` or `maxLines="2"`, then there should be some place where user can find whole topic name. (as in topic narrow it is ellipsize)
cb95a3ed-7c2c-4db1-9935-fc25e89efe65

@Sam1301 What is the screen size of the [above image](https://github.com/zulip/zulip-android/pull/437#issuecomment-296373509)?

Sam1301:
@vishwesh3 ldpi (320 X 426 in dp)

vishwesh3:
One solution :- we can limit stream, topic name to `ellipsize="end"` and `maxLines="1"` or `maxLines="2"` . And on long press (stream or topic) we can show full name in the header itself or in a toast.

Thoughts?

kunall17:
Yeah, i think maxlines = 1 would be fine (2 would be too much packed up) 
We can use the click there in the toolbar as noticing long press in the toolbar is a little difficult and no one would expect that, on click there can be more possibility?

vishwesh3:
Yes, we can also make toolbar clickable.

I was thinking, long press on message header.

vishwesh3:
Like this
ezgif com-video-to-gif 29

divs4debu:
I think that would be too much of user interaction on message header.

divs4debu:
What u can do is that when the header reaches the top it can show full topic and that might look good too.

kunall17:
@vishwesh3 This interaction seems nice, but what about the topic header as Saumya mentioned, that would create problem, but otherwise this seems good.

I would like to wrap up this PR with this interaction, so rebase and we'll do a release! :)
And if you're covering the toolbar click handler too do them in different commits! 

@divs4debu If we do something like that, while scrolling the header will be constantly getting big and small which would not look good!

vishwesh3:
rebased

kunall17:
Awesome work, merged thanks @vishwesh3 👍

Issue #277, opened 2017-01-02
Add support for emoji reactions


timabbott:
We recently merged into master support for emoji reactions to messages in Zulip.  We need to do a few things:
- [ ] Display existing emoji reactions after loading message history
- [ ] Handle real-time sync of new emoji reactions being added by other users
- [ ] Add a long-press menu item to add an emoji reaction

The web frontend implementation (in master but not quite yet on chat.zulip.org) is a useful guide.

saketkumar95:
@timabbott I'm currently done with 70% of work in this issue https://github.com/zulip/zulip-android/issues/248
What else do i need to do more??

Sam1301:
This seems to be an interesting task :) I'd like to work on it if no one else is assigned yet.

timabbott:
Let me deploy the new feature to chat.zulip.org; that will make it easier to work with.

anuradhagupta345:
@saketkumar95 What is your progress with this issue?

Sam1301:
Hey @anuradhagupta345, I'll be working on issue.

PR #451, opened 2017-03-07
Display Reactions

Sam1301:
Addresses first 2 tasks of #277 

**Summary of changes**
* render reactions
* listen for reaction type events

screenshot_20170307-070135

Brief video of reaction events in action :D
https://drive.google.com/open?id=0B4jK5QX65b0ZNVZjZ1Z0LTlFYU0

- [x] Code is formatted.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
+++ b/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
@@ -442,32 +442,55 @@ public class RecyclerMessageAdapter extends RecyclerView.Adapter<RecyclerView.Vi
                         messageHolder.reactionsTable.removeAllViews();
                         return;
                     }
-                    /**
-                     * Calculate number of reactions per row
-                     */
-                // TODO: use when using multiple rows
-                    int numOfReactions = messageHolder.reactionsTable.getWidth() / context.getResources().getDimensionPixelSize(R.dimen.reaction_background_w);
+
+                    // Calculate number of reactions in each row
+                    int messageWidth = messageHolder.messageTile.getContext().getResources().getDisplayMetrics().widthPixels;
+                    int reactionWidth = convertDpToPixel(64);
+                    int numOfReactions = messageWidth / reactionWidth;
                     TableRow row = new TableRow(messageHolder.messageTile.getContext());
+
+                    // set margin of 6dp between reactions in a table row
                     TableRow.LayoutParams layoutParams = new TableRow.LayoutParams();
-                    int margin = ConvertDpPx.convertDpToPixel(4);
+                    int margin = ConvertDpPx.convertDpToPixel(6);
                     layoutParams.setMargins(margin, margin, margin, margin);
+
+                    // table row index
+                    int index = 0;
                     for (Reaction reaction : message.getReactions()) {
+                        if (numOfReactions == 0) {
+                            // current row is full
+                            messageHolder.reactionsTable.addView(row, index++);
+                            row = new TableRow(messageHolder.messageTile.getContext());
+                            numOfReactions = messageWidth / reactionWidth;
+                        }
+
+                        // inflate reaction layout
                         LinearLayout reactionTile = (LinearLayout) LayoutInflater.from(messageHolder.messageTile.getContext()).inflate(R.layout.reaction_tile, null);
                         reactionTile.setLayoutParams(layoutParams);
+
+                        // emoji view in reaction
                         ImageView imageView = (ImageView) reactionTile.findViewById(R.id.reaction_emoji);
+                        // emoji count view
                         TextView textView = (TextView) reactionTile.findViewById(R.id.reaction_count);
+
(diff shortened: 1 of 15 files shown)


timabbott:
Nice!!

timabbott:
We should make sure this code works against a server that doesn't support reactions yet (e.g. running 1.4.x).

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
```diff
@@ -427,6 +439,64 @@ public void onClick(View view) {
...
                 setUpStar(message, messageHolder);
+                try {
+                    messageHolder.reactionsTable.removeAllViews();
+                    if (message.getReactions().isEmpty()) {
+                        return;
+                    }
+
+                    // Calculate number of reactions in each row
+                    int messageWidth = messageHolder.messageTile.getContext().getResources().getDisplayMetrics().widthPixels;
+                    int reactionWidth = convertDpToPixel(64);
```
kunall17: Better to save this 64 in Constants

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
```diff
@@ -427,6 +439,64 @@ public void onClick(View view) {
...
+                    int numOfReactions = messageWidth / reactionWidth;
+                    TableRow row = new TableRow(messageHolder.messageTile.getContext());
```
kunall17: We can use the `context` object here

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
```diff
@@ -427,6 +439,64 @@ public void onClick(View view) {
...
+
+                    // set margin of 6dp between reactions in a table row
+                    TableRow.LayoutParams layoutParams = new TableRow.LayoutParams();
+                    int margin = ConvertDpPx.convertDpToPixel(6);
```
kunall17: It would be better to move all this in its own method setUpReactions

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
```diff
@@ -427,6 +439,64 @@ public void onClick(View view) {
                 setUpGravatar(message, messageHolder);
                 setUpTime(message, messageHolder);
                 setUpStar(message, messageHolder);
+                try {
+                    messageHolder.reactionsTable.removeAllViews();
```
kunall17: Why `removeAllViews` is needed here?

app/src/main/java/com/zulip/android/models/User.java
```diff
@@ -0,0 +1,68 @@
+package com.zulip.android.models;
+
+import com.google.gson.annotations.SerializedName;
+import com.j256.ormlite.field.DatabaseField;
+import com.j256.ormlite.table.DatabaseTable;
+
+@DatabaseTable(tableName = "reactions_user")
+public class User {
```
kunall17: Please rename this to something `UserReaction`

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
Sam1301: `recylerview` does a perf optimisation where views are reused instead of creating new ones. Hence to avoid any erroneous display of reactions on a message tile, the reactions property is reset from removing all views from the reactions table.

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
Sam1301: Not sure if I got your point here. Do you mean this [context](https://github.com/Sam1301/zulip-android/blob/ede40c540d898ab39e999c7ddf15f2e232892270/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L114) ?
I intended to pass the messagetile context since new Table rows need to be create within a single messagetile.
According to the docs,
```
/**
     * <p>Creates a new TableRow for the given context and with the
     * specified set attributes.</p>
     *
     * @param context the application environment
     * @param attrs a collection of attributes
     */
    public TableRow(Context context, AttributeSet attrs) {
        super(context, attrs);
        initTableRow();
    }
```

app/src/main/java/com/zulip/android/models/User.java
Sam1301: 👍

Sam1301:
@kunall17 updated pr.

kunall17:
@Sam1301 there's some problem with the commit author details, you might want to check it out! 

But I think when i'll rebase and merge this into master it will be overridden to `sam1301 committed with kunall17` so its not gonna make difference!

kunall17:
Thanks @Sam1301, there was a small conflict in the Constants.java i fixed that, and you forgot to increase the database version which I did and merged! 👍

Issue #226, opened 2016-11-16
App not updating with latest messages


niftynei:
A few users have reported that the newest messages are failing to load.  Killing the app and restarting seems to fix.

Sam1301:
@niftynei  I tested this in the beta and release candidate and it seemed to work. The newest messages are added at the bottom of homeView, though the users have to scroll down to them (without refreshing).

niftynei:
yes, it seems to work in most cases, but there appears to be a hard to replicate state where messages are no longer refreshed.  i managed to get my app into this state, but i don't know how to reproduce reliably.

bengoldstein:
I'm using the Play Store edition release March 1 , 2017, and having this problem. By "this problem" I mean that the posts I make in a topic frequently do not show up unless I refresh. (I do see what other users post. Also, I do see what I post from my Android in my desktop client.)

It *may* have something to do with my using both the desktop client and the Android app at the same time. I'm not able to reliably reproduce this problem either, although it is happening a lot for me--basically making the app unusable. (I have tried this on two cell phones--a Samsung Galaxy S5 and a Google Nexus 5X with the same results.)

PR #477, opened 2017-03-24
Fixed topic indicator/separator showing for new messages when user has the topic open

### LINKED ISSUE #441: remove topic indicator for new messages in same topic
**Actual Behaviour**
This topic indicator/separator appears even when user is already in that topic (not in home) and replies to a message. That gives the impression that one is still in home thread or the text above the indication belongs to a different topic. 
Also; three times "android review" on the same page doesn't need to be. 
20170304_22-00-58_paintdotnet

**Expected Behaviour**
The indicator doesn't show up for new messages when user has one topic opened and replies to it.

prati0100:
Fixes #441

**Summary of changes**

The for loop in `addNewMessage()` in `RecyclerMessageAdapter.java` goes from `getItemCount(false)` to 1, but not equal to 1. When the user has opened a topic, the only `MessageHeaderParent` in the list is at the top of the list of items. When i does not go to 1, the `if(items.get(i) instanceof MessageHeaderParent)` never gets executed. This means `item` remains null and `if(item == null)` block is executed, resulting in the creation of the `MessageHeaderParent` above the new message, which is not wanted.

By making it `i >= 1`, the `MessageHeaderParent` at the top of the list is also included.

- [X] Code is formatted.

- [X] Run the lint tests with `./gradlew lintDebug` and make sure BUILD is SUCCESSFUL

- [X] If new feature is implemeted then it is compatible with Night theme too.

- [X] Commit messages are well-written.

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
+++ b/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
@@ -289,7 +289,7 @@ public class RecyclerMessageAdapter extends RecyclerView.Adapter<RecyclerView.Vi
      */
     public void addNewMessage(Message message) {
         MessageHeaderParent item = null;
-        for (int i = getItemCount(false) - 1; i > 1; i--) {
+        for (int i = getItemCount(false) - 1; i >= 1; i--) {
             //Find the last header and check if it belongs to this message!
             if (items.get(i) instanceof MessageHeaderParent) {
                 item = (MessageHeaderParent) items.get(i);


smarx:

@prati0100, it looks like you've already signed the Dropbox CLA. Thanks!

app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java
```diff
@@ -289,7 +289,7 @@ public boolean addOldMessage(Message message, int messageAndHeadersCount, String
      */
     public void addNewMessage(Message message) {
         MessageHeaderParent item = null;
-        for (int i = getItemCount(false) - 1; i > 1; i--) {
+        for (int i = getItemCount(false) - 1; i >= 1; i--) {
```
prati0100: Why not?

Firstly, I tried the code and it works fine for me. The messages are added to the bottom and I couldn't find any bug during my testing.

And even in principle, I don't see why messages would not be added at the bottom of the list. The messages are added using `items.add(getItemCount(true) - 1, message)`, which does not use the value of i anywhere. I don't know why one more iteration of the for loop would mess with the message adding part. Earlier, the MessageHeaderParent at the top wasn't accounted for. With i going equal to 1, that is fixed.

I don't know why you think this is a problem. Did you try running the code? If you are seeing something wrong when testing, please let me know, and I'll try on other devices. At least on my device, the code works fine.

Sam1301: you're correct; when I tried the changes, new messages were not being loaded (might be related to #226) 

double checked the logic; works well 👍

Sam1301:
@prati0100 Can you please clean up the commit message to follow our [commit message style](http://zulip.readthedocs.io/en/latest/version-control.html#commit-messages_)?

prati0100:
I don't get why @zulipbot is saying that I have not referenced the issue in my commit message. If you read the commit message, I have clearly written "Fixes #441". Should that be in the first line?

Also, @Sam1301 would you suggest any other changes to the commit message?

kunall17:
Awesome, works well thanks for working on this! Merged @prati0100 👍

PR #480, opened 2017-03-29
Revert "Sorted people list in people drawer."

kunall17:
@vishwesh3 The changes you made were really nice, but after some usage of the latest master I some times faced slight jitters + lag (often when doing narrows and stuff)
So I tried to do some profiling 

With sorting (local env)
non revert local dev

Without sorting local dev
revert local dev

With sorting (Chat.zulip.org)
non revert chat zulip

Without sorting (Chat.zulip.org)
revert chat zulip

Because persisting of the PeopleDrawerList.java model in a list for the recyclerView adapter seems to be the problem, and the memory usage is huge as well considering the lower end phones would definitely feel some glitches! 

Also Reverting is a painful thing and I did a quick fix, so maybe we can give this branch a test if anything is broke, (After deciding we should revert or not) 
Need your review @Sam1301 @saketkumar95 @nitish1211 @yadav-rahul @abhaymaniyar


smarx:

kunall17:
Merged!

PR #497, opened 2017-05-14
Fix: Use unicode to locate emojis.

Sam1301:
Old code relied on the `title` attribute to locate emojis.
In case of unicode emojis, the `title` attribute gives the unicode character itself.

To remove this ambiguity, the new code uses cssClass to locate the emoji file for all emojis.


kunall17:
Merged, thanks @Sam1301 but please consider doing different commits for fixing lint issues! :)

Sam1301:
Ah missed that! Will keep in mind :)

PR #498, opened 2017-05-16
fix: app crash on sending PM notification bouncer.

vishwesh3:
**Summary of changes**

 url does not contain realm url (and protocol) so it was crashing. In this PR :- join ServerHostUri and url to get full url of sender avatar.

https://chat.zulip.org/user_uploads/2/ea/YX_itZYiV0OCGazCMqbM3xVo/Screenshot_1494931543.png
https://chat.zulip.org/user_uploads/2/7c/PSAvceGaV2g_hvZYf3xXd7d7/18552755_796366293890514_954585702_o.jpg

- [x] Code is formatted.

### DIFF
--- a/app/src/main/java/com/zulip/android/networking/GravatarAsyncFetchTask.java
+++ b/app/src/main/java/com/zulip/android/networking/GravatarAsyncFetchTask.java
@@ -15,6 +15,7 @@ import android.widget.ImageView;
 
 import com.zulip.android.activities.ZulipActivity;
 import com.zulip.android.models.Person;
+import com.zulip.android.util.UrlHelper;
 import com.zulip.android.util.ZLog;
 
 import java.io.IOException;
@@ -47,7 +48,7 @@ public class GravatarAsyncFetchTask extends AsyncTask<URL, Void, Bitmap> {
         float px = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP,
                 dpSize, r.getDisplayMetrics());
         try {
-            return new URL(url + "&s=" + px);
+            return new URL(UrlHelper.addHost(url)  + "&s=" + px);
         } catch (MalformedURLException e) {
             ZLog.logException(e);
             return null;
--- a/app/src/main/java/com/zulip/android/util/ListDialog.java
+++ b/app/src/main/java/com/zulip/android/util/ListDialog.java
@@ -2,12 +2,14 @@ package com.zulip.android.util;
 
 import android.app.AlertDialog;
 import android.app.Dialog;
+import android.support.annotation.NonNull;
 import android.support.v4.app.DialogFragment;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
+import android.view.ViewGroup;
 import android.widget.ImageView;
 
 import com.zulip.android.R;
@@ -40,6 +42,7 @@ public class ListDialog extends DialogFragment {
         }
     }
 
+    @NonNull
     @Override
     public Dialog onCreateDialog(Bundle savedInstanceState) {
         AlertDialog.Builder builder = new AlertDialog.Builder(getActivity());
@@ -48,7 +51,8 @@ public class ListDialog extends DialogFragment {
 
         // Inflate and set the layout for the dialog
         // Pass null as the parent view because its going in the dialog layout
-        View rootView = inflater.inflate(R.layout.list_dialog, null);
+        final ViewGroup nullParent = null;
+        View rootView = inflater.inflate(R.layout.list_dialog, nullParent);
         View cameraListItem = rootView.findViewById(R.id.picture_dialog);
         View fileListItem = rootView.findViewById(R.id.pick_file_dialog);
         ImageView cameraImage = (ImageView) rootView.findViewById(R.id.camera_share_icon);


app/src/main/java/com/zulip/android/networking/GravatarAsyncFetchTask.java
```diff
@@ -47,7 +48,7 @@ public static URL sizedURL(Context context, String url, float dpSize) {
         float px = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP,
                 dpSize, r.getDisplayMetrics());
         try {
-            return new URL(url + "&s=" + px);
+            return new URL(ZulipApp.get().getServerHostUri() + url + "&s=" + px);
```
timabbott: Just to confirm: Zulip sends both relative URLs (for custom avatars) and absolute URLs (for default gravatars); does this code handle both cases?

vishwesh3: Updated :+1: Now it handles both cases.
18575338_796557247204752_1421320719_o
18519306_796557150538095_1647942643_o

timabbott:
@Sam1301 @kunall17 can you give this a quick review?  I'm thinking we should release a 1.3.2 release in the next couple days with just this change and the emoji fix in #497

app/src/main/java/com/zulip/android/networking/GravatarAsyncFetchTask.java
```diff
@@ -47,7 +48,7 @@ public static URL sizedURL(Context context, String url, float dpSize) {
         float px = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP,
                 dpSize, r.getDisplayMetrics());
         try {
-            return new URL(url + "&s=" + px);
+            return new URL(UrlHelper.getSenderAvatarUrl(url)  + "&s=" + px);
```
Sam1301: I think we could simply use `addHost` method of UrlHelper (see `RecyclerMessageAdapter.setUpGravatar`).

app/src/main/java/com/zulip/android/util/ListDialog.java
```diff
@@ -48,7 +51,8 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
 
         // Inflate and set the layout for the dialog
         // Pass null as the parent view because its going in the dialog layout
-        View rootView = inflater.inflate(R.layout.list_dialog, null);
+        final ViewGroup nullParent = null;
+        View rootView = inflater.inflate(R.layout.list_dialog, nullParent);
```
Sam1301: any specific reason behind this change?

vishwesh3: just lint fix

Sam1301: I see, how about just suppressing this error since we are using `null` again as the view root here.

app/src/main/java/com/zulip/android/networking/GravatarAsyncFetchTask.java
Sam1301: just to clarify: using new `getSenderAvatarUrl` func is redundant as it does probably the same job as UrlHelper.addHost
[RecyclerMessageAdapter.setUpGravatar](https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/activities/RecyclerMessageAdapter.java#L788) is where we use addHost for the same purpose.

vishwesh3: Yeah! I missed that. Thanks. Updated :+1:

app/src/main/java/com/zulip/android/networking/GravatarAsyncFetchTask.java
```diff
@@ -47,7 +48,7 @@ public static URL sizedURL(Context context, String url, float dpSize) {
         float px = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP,
                 dpSize, r.getDisplayMetrics());
         try {
-            return new URL(url + "&s=" + px);
+            return new URL(UrlHelper.addHost(url)  + "&s=" + px);
```
kunall17: @timabbott Doing this change can mess up things for the GCM for zulipchat servers as well, so now will the servers not use GCM and use push bouncer? 

(I never got to work with GCM so I'm not sure what `sender_avatar_url` string contains!

kunall17: @vishwesh3 It would be better to have a check the URL is incomplete then only add the host in there. It would be great if you could do it really soon, want to get this working really soon :)

vishwesh3: yes we are checking whether url starts with http or not in the `addHost` method.
https://github.com/zulip/zulip-android/blob/master/app/src/main/java/com/zulip/android/util/UrlHelper.java#L8

Do we want any other checks?

kunall17: Ok yeah, didn't thought that addHost would also check for the hosts! cool 👍

kunall17:
@vishwesh3 you might want to change the author details its coming as `vishwesh3 commited with vishwesh3` ?
Or should I push them right away?

vishwesh3:
@kunall17 You can push them. 👍

kunall17:
Thanks for working on this @vishwesh3, merged :)

PR #504, opened 2017-05-23
Fixes #503: Fix the camera opening on clicking the notification

### LINKED ISSUE #503: Camera opening on clicking the message push notification

kunall17:
So it seems the action in the intent from the notification has the action as null and its going to `dispatchTakePictureIntent();` as the filepath will be null as well, so the camera app is opening on pressing notification!

### DIFF
--- a/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
+++ b/app/src/main/java/com/zulip/android/activities/ZulipActivity.java
@@ -141,6 +141,7 @@ import okhttp3.MultipartBody;
 import retrofit2.Call;
 import retrofit2.Response;
 
+import static com.zulip.android.util.Constants.ACTION_MESSAGE_PUSH_NOTIFICATION;
 import static com.zulip.android.util.Constants.REQUEST_PICK_FILE;
 
 /**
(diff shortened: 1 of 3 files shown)


smarx:

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -893,6 +894,8 @@ protected void onNewIntent(Intent intent) {
                 // photo was deleted and camera is launched again to capture a new photo
                 dispatchTakePictureIntent();
             }
+        } else if (action.equals(ACTION_MESSAGE_PUSH_NOTIFICATION)) {
```
kunall17: I am aware this thing is not required, but lets have.
It would help the developers to use the intent if they want to!

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -893,6 +894,8 @@ protected void onNewIntent(Intent intent) {
...
+            //Do nothing here
```
timabbott: explain in the comment why you should do nothing here.

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
Sam1301: cool, structurally I think it might be better to handle this case in the preceding if-else as that is for discerning actions. Works well though 👍

Issue #505, opened 2017-06-02
Theme is lost when closing/reopening the app


timabbott:
This is from beta feedback on the app:

Bug: If you select a dark theme and close the application, a light theme will be displayed when you open it.

vishwesh3:
@zulipbot claim

zulipbot:
Hello @vishwesh3, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 7 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #506, opened 2017-06-03
Restore theme after closing/reopening the app.

vishwesh3:
Fix:#505

**Summary of changes**

Save preferred theme in Shared Preferences. And restore them on reopening app.

ezgif com-video-to-gif-2

- [x] Code is formatted.

### DIFF
--- a/app/src/main/java/com/zulip/android/ZulipApp.java
+++ b/app/src/main/java/com/zulip/android/ZulipApp.java
@@ -598,4 +598,15 @@ public class ZulipApp extends Application {
             }
         }
     }
+
+    /**
+     * Update preferred theme
+     * @param b true if night theme is preferred
+     */
+    public void makeNightThemeDefault(boolean nightTheme) {
+        SharedPreferences.Editor editor = getSettings().edit();
+
+        editor.putBoolean(Constants.NIGHT_THEME, nightTheme);
+        editor.apply();
+    }
 }
--- a/app/src/main/java/com/zulip/android/ZulipApp.java
+++ b/app/src/main/java/com/zulip/android/ZulipApp.java
@@ -604,6 +604,7 @@ public class ZulipApp extends Application {
 
     /**
      * Update preferred theme
+     *
      * @param b true if night theme is preferred
      */
     public void makeNightThemeDefault(boolean nightTheme) {
@@ -615,6 +616,7 @@ public class ZulipApp extends Application {
 
     /**
      * update auto night theme preference
+     *
      * @param value preference value
      */
     public void setAutoNightTheme(boolean value) {
--- a/app/src/main/java/com/zulip/android/ZulipApp.java
+++ b/app/src/main/java/com/zulip/android/ZulipApp.java
@@ -7,6 +7,7 @@ import android.content.SharedPreferences.Editor;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager.NameNotFoundException;
 import android.os.Handler;
+import android.support.v7.app.AppCompatDelegate;
 import android.util.Log;
 
 import com.crashlytics.android.Crashlytics;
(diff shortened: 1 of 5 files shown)


kunall17:
@timabbott @vishwesh3 @Sam1301 should we use the auto night theme, it changes the theme according to the time going on?

https://developer.android.com/reference/android/support/v7/app/AppCompatDelegate.html#MODE_NIGHT_AUTO

Sam1301:
Nice idea!, though I am somewhat biased to user having control over the theme regardless of time, but this is debatable.

saketkumar95:
I guess No! We should keep the preference of the User only. Twitter app flows the same flow, i.e It all depends on the user choice. But It's nice idea. Some apps like Google maps uses this. :) 
@kunall17 Lets checkout what Slack's follows..  :D

kunall17:
Yeah but imagine a situtation where I enabled the night theme around 9 PM, and then I started the app in the next morning I would prefer, that it would change automatically to the day mode in the morning! 

One solution is we can have a sub menu like the Filter menu, and have a checkbox there for auto changing and one option for changing the current theme?

app/src/main/java/com/zulip/android/ZulipApp.java
```diff
@@ -598,4 +598,15 @@ public void setMessageContentEditParams(int seconds, boolean param) {
             }
         }
     }
+
+    /**
+     * Update preferred theme
+     * @param b true if night theme is preferred
+     */
+    public void makeNightThemeDefault(boolean b) {
```
kunall17: Make this boolean b something more understandable!

app/src/main/java/com/zulip/android/util/Constants.java
```diff
@@ -27,4 +27,5 @@
     //if two continuous messages are from same sender and time difference is less than this then hide it
     public static long HIDE_TIMESTAMP_THRESHOLD = 60 * 1000;// 1 minute
     public static String ACTION_MESSAGE_PUSH_NOTIFICATION = "ACTION_MESSAGE_PUSH_NOTIFICATION";
+    public static String NIGHT_THEME = "isNightTheme";
```
kunall17: Let's follow the convention going on, `public static String NIGHT_THEME = 'NIGHT_THEME'`

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
```diff
@@ -2358,12 +2365,15 @@ public void onClick(
                     case -1:
                     case AppCompatDelegate.MODE_NIGHT_NO:
                         setNightMode(AppCompatDelegate.MODE_NIGHT_YES);
+                        app.makeNightThemeDefault(true);
                         break;
                     case AppCompatDelegate.MODE_NIGHT_YES:
```
kunall17: NIT: As the default and this is same, we can remove the block (Upto you how you want to do it) 
```
setNightMode(AppCompatDelegate.MODE_NIGHT_NO);
app.makeNightThemeDefault(false);
```

app/src/main/java/com/zulip/android/ZulipApp.java
vishwesh3: Done 👍

app/src/main/java/com/zulip/android/util/Constants.java
vishwesh3: Done 👍

app/src/main/java/com/zulip/android/activities/ZulipActivity.java
vishwesh3: Done 👍 . Removed block.

vishwesh3:
Updated 👍 

Added functionality for auto night theme.
- Added a checkbox option in menu.
- If auto theme is enabled then irrespective of preffered theme, it
will be overridden by auto theme (depending on time).
- User can switch theme even if auto theme is enabled, but it is for that instance
only.

kunall17:
Thanks for the quick awesome work @vishwesh3.
Waiting for confirmation of @timabbott to merge this PR, after that we can do a beta release again.

timabbott:
I think this seems reasonable.  One the question of whether to do auto-night, I think long-term maybe we want 3 setting values for this sort of feature: day/night/auto, so the user can choose how they want it to work.

vishwesh3:
@timabbott Added 3 options.
3464974478807180902 account_id 1
3757914563653689949 account_id 1

Sam1301:
nice! we can probably omit "theme" in each option.

vishwesh3:
@Sam1301 Updated 👍 
screenshot_1496639437

kunall17:
Cool this looks awesome! 👍

kunall17:
Merged #506 👍

Issue #405, opened 2017-02-17
Make sure that all of our strings have the standard Zulip capitalization scheme


timabbott:
Generally, just the first letter of phrases is capitalized in Zulip, even in headings/etc.  E.g.
```
Search Messages -> Search messages
Find Users -> Find users
```

saketkumar95:
yeah, i will check all the strings, if found such then do the changes. :)

saketkumar95:
@timabbott what about the string "Welcome to Zulip!" ? should i change that too?

Sam1301:
@saketkumar95 I guess not, proper nouns can be left capitalized.

timabbott:
Yeah, capital nouns should be left capitalized as normal :).

PR #510, opened 2017-06-30
Improve strings.

YagoGG:
**Summary of changes**

Strings in `strings.xml` have been properly capitalized, following the convention in #405.

"Terms of Service" has been left with "Service" capitalized, as many other services do (like [Google's TOS](https://www.google.com/accounts/TOS)).

**Other considerations**

I'd suggest that someone more familiar with the app's codebase reviewed:

- `<string name="link_image">ImageView of link</string>`
- `<string name="crop">CROP</string>`
- `<string name="placeholder_my_stream">MyStream</string>`

As those seem to be strings that cannot be changed without some modifications to the code that uses them. The first and the third one shouldn't be translated (or at least in that form).

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -8,7 +8,7 @@
     <string name="password">Password</string>
     <string name="server_url">Server URL</string>
     <string name="server_domain">Server domain</string>
-    <string name="privacy">Privacy Policy</string>
+    <string name="privacy">Privacy policy</string>
     <string name="logout">Log out</string>
     <string name="force_logged_out">You\'ve been logged out</string>
     <string name="huddle_text">You</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -175,11 +175,13 @@
...
+    <string name="placeholder_short_stream">Short stream</string>
     <string name="placeholder_topic">Topic</string>
     <string name="placeholder_message_here">Message here</string>
     <string name="placeholder_sender_name">Sender name</string>
     <string name="placeholder_social">Social</string>
     <!--Placeholder strings-->
-    <string name="terms">Terms Of Service</string>
+    <string name="terms">Terms of Service</string>
+
+
```
saketkumar95: remove this extra lines

YagoGG:
@saketkumar95 Thanks for reviewing! I just removed the newlines, this should be ready now :)

kunall17:
Merged #510 👍

PR #518, opened 2018-08-12
deprecation: Improve the deprecation notice

borisyankov:
Makes the app notice more user-centric. We are making it more clear
this app will soon no longer be available.

### DIFF
--- a/app/src/main/java/com/zulip/android/util/ShowAppUpdateDialog.java
+++ b/app/src/main/java/com/zulip/android/util/ShowAppUpdateDialog.java
@@ -44,8 +44,8 @@ public class ShowAppUpdateDialog {
         });
 
         AlertDialog dialog = builder.create();
-        dialog.setMessage("This project is deprecated in favor of React Native app, " +
-                "which supports both iOS and Android. Click update to switch to the new app.");
+        dialog.setMessage("This app is no longer supported and will soon be removed from the Play Store.\n"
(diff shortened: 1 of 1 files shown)


gnprice:
This text looks good! I have a few suggestions for further improvement in this interaction:
* Let's say something about how to get in touch with us; perhaps "file an issue on github.com/zulip/zulip-mobile".
* For the onclick handler that is meant to link to the Play Store, let's follow the recommendation [in upstream docs](https://developer.android.com/distribute/marketing-tools/linking-to-google-play):
   * Follow the URL pattern there -- so no `hl` query-param.
   * Add `intent.setPackage("com.android.vending")`, so it actually goes straight to the Play Store app.
* For the text of the buttons:
   * "Ignore" rather than "Cancel" -- this is something we're bugging them about that they can ignore for now, not something they've requested.
   * "Update" isn't good because that's the normal term for getting a new version of *this* app. Better would be something about "install", perhaps "Install new app".
   * I'm dubious of the "ignore permanently" logic that the "Don't show again" button provides -- there's basically no scenario where that's a good experience, compared to reminding again in a week or a month. But the total number of users still on this app may be few enough it's not worth adding new logic for that.

Also the CI failed, saying something about lint. I haven't tried to look into it, and haven't worked with this codebase so I don't have any context on that, but hopefully it's easy to just fix.

